PR #23, opened 2024-10-16
Update README.md

### LINKED ISSUE #19: Update project README
Update the project `README.md` with the following components:

- Status badge for `CI` (`main.yml`) workflow
- **Motivation** section explaining the motivation behind the project in 2-3 sentences
- **Usage** / **Examples** section giving basic hello world example showing how the semantic analyzer & validator work
- **Community** / **Contributing** section outlining the various ways others can get involved in the project

TheGupta2012:
Fixes #19 
## Summary of changes
- Update `README.md` with project info and examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ pip install pyqasm
 You can also install from source by cloning this repository and running a pip install command
 in the root directory of the repository:
 
-```shell
+```bash
 git clone https://github.com/qBraid/pyqasm.git
 cd pyqasm
 pip install .
@@ -52,7 +52,7 @@ print(pyqasm.__version__)
 ### Unrolling OpenQASM 3 program 
 
 ```python
-from pyqasm.unroller import unroll
+import pyqasm 
 
 program = """
 OPENQASM 3;
@@ -64,10 +64,10 @@ bit[2] c;
 h q[0];
 cx q[0], q[1];
 
-measure q->c;
+c = measure q;
 """
 
-unrolled_qasm = unroll(program).unrolled_qasm
+unrolled_qasm = pyqasm.unroll(program)
 print(unrolled_qasm)
 ```
 
@@ -76,7 +76,7 @@ For a more complex example, see the [Deutsch Josza program unrolling](examples/u
 ### Validating OpenQASM 3 program 
 
 ```python
-from pyqasm.validate import validate
+import pyqasm
 
 program = """
 OPENQASM 3;
@@ -91,7 +91,7 @@ cx q[0], q[1];
 measure q -> c;
 """
 
-assert validate(program) is None
+pyqasm.validate(program)
 ```
 `validate` returns None if the program is semantically valid, otherwise raises an Exception. See the [validation example](examples/validate_example.py) for more insight into the capabilities of our analyser.
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,13 @@
 
 Python toolkit providing an OpenQASM 3 semantic analyzer and utilities for program analysis and compilation.
 
+
+>[!WARNING]
+> **This project is "pre-alpha", and is not yet stable or fully realized. Use with caution, as the API and functionality are subject to significant changes.**
+
+## Motivation 
+The current [OpenQASM 3 standard](https://openqasm.com/index.html) is a powerful language for expressing hybrid quantum-classical programs, but it lacks a comprehensive tool supporting the full capabilities of the language. Pyqasm aims to fill this gap by building upon the [openqasm parser](https://github.com/openqasm/openqasm/tree/main/source/openqasm), and providing support for semantic analysis and utilities for program compilation.
+
 ## Installation
 
 pyqasm requires Python 3.10 or greater, and can be installed with pip as follows:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -6,10 +6,6 @@
     <img src='https://github.com/qBraid/pyqasm/actions/workflows/main.yml/badge.svg' alt='CI'>
 </a>
```
ryanhill1, line 8: Use status badge in this format to be consistent the the ones below:

```
[![CI](https://github.com/qBraid/pyqasm/actions/workflows/main.yml/badge.svg?branch=main)](https://github.com/qBraid/pyqasm/actions/workflows/main.yml)
```

I only used HTML syntax for the pyqasm logo so that I could use right align

README.md
```diff
@@ -26,5 +29,87 @@ pyqasm requires Python 3.10 or greater, and can be installed with pip as follows
...
 ```
 
->[!WARNING]
-> **This project is "pre-alpha", and is not yet stable or fully realized. Use with caution, as the API and functionality are subject to significant changes.**
+## Install from source 
+
+You can also install from source by cloning this repository and running a pip install command
+in the root directory of the repository:
+
+```shell
```
ryanhill1, line 37:
```suggestion
```bash
```

to be consistent

README.md
```diff
@@ -26,5 +29,87 @@ pyqasm requires Python 3.10 or greater, and can be installed with pip as follows
...
+
+h q[0];
+cx q[0], q[1];
+
+measure q->c;
+"""
+
+unrolled_qasm = unroll(program).unrolled_qasm
+print(unrolled_qasm)
+```
```
ryanhill1, line 74: Let's write import like this, a little cleaner
 
```python
import pyqasm

...

unrolled_qasm = pyqasm.unroll(program)
```

Also, can we make it so that `pyqasm.unroll` returns the unrolled qasm string directly intead of a `Qasm3Module`? Unless there's a good reason to keep it in that object.

Also, correct me if I'm wrong, but I think that

```
measure q -> c;
```

is OpenQASM 2 syntax, and is still just supported by OpenQASM 3 for backwards compatibility. I'm pretty sure the new syntax is something like

```
c[0] = measure q[0]; // single qubit
c = measure q;  // multiple qubits
```

I know that's correct for single, but might wanna double check for multiple

README.md
```diff
@@ -26,5 +29,87 @@ pyqasm requires Python 3.10 or greater, and can be installed with pip as follows
...
+
+For a more complex example, see the [Deutsch Josza program unrolling](examples/unroll_example.py) 
+
+### Validating OpenQASM 3 program 
+
+```python
+from pyqasm.validate import validate
```
ryanhill1, line 81:
```suggestion
import pyqasm
```

README.md
```diff
@@ -26,5 +29,87 @@ pyqasm requires Python 3.10 or greater, and can be installed with pip as follows
...
+bit[2] c;
+// create a Bell state
+h q[0];
+cx q[0], q[1];
+
+// measure the qubits
+measure q -> c;
+"""
+
+assert validate(program) is None
```
ryanhill1, line 96:
```suggestion
pyqasm.validate(program)
```

examples/unroll_example.py
```diff
@@ -53,4 +53,5 @@ def deutsch_jozsa(qubit[N] q_func, qubit[1] ancilla_q) {
 measure q -> result;
 """
 
-print(unroll(example_qasm_program).unrolled_qasm)
\ No newline at end of file
+print(unroll(example_qasm_program).unrolled_qasm)
```
ryanhill1, line 56: Split into two lines

examples/validate_example.py
```diff
@@ -0,0 +1,70 @@
+from pyqasm.validate import validate
```
ryanhill1, line 1:
```suggestion
import pyqasm
```

examples/validate_example.py
```diff
@@ -0,0 +1,70 @@
...
+// Apply parameterized rotations
+for int i in [0:N-1] {
+    rx(angles[i]) q[i];
+}
+
+// Measure the qubits
+measure q -> c;
+"""
+
+assert validate(qasm_program) is None
```
ryanhill1, line 70:
```suggestion
pyqasm.validate(program)
```

ryanhill1 (changes requested):

README.md
TheGupta2012: Yea so `qbraid-qir` currently expects an `openqasm3.ast.Program` object which is present as one of the attributes of `Qasm3Module`. We can have an arg like - 
```python
def unroll(program, return_type : str = "str"):
   # return_type is one of {"str", "ast"} 
``` 
which just returns the repr. 

The only thing to think about is that qir programs require the total num of [qubits and clbits to be declared at the start](https://github.com/qBraid/qbraid-qir/blob/4b0df921e8831bf31d2d0630da52e686f20adb18/qbraid_qir/qasm3/visitor.py#L105) and if we just return the repr, we lose that info and would likely have to calculate that again.

TheGupta2012: Also, the measure format `c = measure q;` works for multiple too, thanks for the suggestion!

ryanhill1: Ok gotcha. Two options in my mind:

1. We could declare a function `pyqasm._unroll()` that returns the `Qasm3Module` that we use internally within qBraid-QIR, and then have `pyqasm.unroll()` be the user facing function that returns the string using `pyqasm._unroll()` as a helper.
2. Add a keyword argument `as_module=False` (or something like that) to the function so that by default, `pyqasm.unroll()` returns a string, by within qBraid-QIR, we can specify `pyqasm.unroll(..., as_module=True)` to keep it in the form of a `Qasm3Module`

Either way works, whichever you think is best

ryanhill1 (approved):

README.md
TheGupta2012: Second one works, will update the code. Thanks!

PR #59, opened 2024-11-05
Support for externally linked gates

Sola85:
This is my preliminary PR for externally linked gates, as suggested here https://github.com/qBraid/qbraid-qir/issues/167.

This is probably not ready to merge, but I figured it's easier if I get some feedback first. I will follow up with a PR for qbraid-qir shortly. With both changes, external gates work in qbraid-qir, at least for my use-case.

With this change `pyqasm` accepts a list of custom gates, that should not be unrolled into their definition.

e.g. with `qasm3_module.unroll(external_gates=["mygate"])`, the following QASM would be unrolled into an AST that contains a "mygate" node, rather than the nodes in the definiton of "mygate".

```qasm
OPENQASM 3.0;
include "stdgates.inc";
gate mygate(p0) _gate_q_0, _gate_q_1 {
  h _gate_q_1;
  cx _gate_q_0, _gate_q_1;
  rz(p0) _gate_q_1;
  cx _gate_q_0, _gate_q_1;
  h _gate_q_1;
}
bit[2] c;
qubit[2] q;
ry(pi/2) q[0];
mygate(pi/2) q[0], q[1];
```


pyqasm/visitor.py
```diff
@@ -759,6 +761,69 @@ def _visit_custom_gate_operation(
...
+                            for more clarity.
+
+        Returns:
+            list[qasm3_ast.QuantumGate]: The quantum gate that was collected.
+        """
+
+        logger.debug("Visiting custom gate operation '%s'", str(operation))
+        gate_name: str = operation.name.name
+
+        if gate_name not in self._custom_gates:
```
TheGupta2012, line 785: I think this should be changed to include the set of basis gates which we support. Eg., if I am calling the `unroll` method like `qasm_module.unroll(external_gates = ["rx"])` and then have an `rx` gate in my program it will raise an error - 
```python 

In [27]: qasm_str = """OPENQASM 3.0;
    ...: include "stdgates.inc";
    ...: qubit[2] q;
    ...: rx(0) q[0];"""

In [28]: mod = pyqasm.load(qasm_str)

In [29]: mod.unroll(external_gates = ["rx"])
ERROR:root:Error at line 4, column 0 in QASM file
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In[29], line 1
----> 1 mod.unroll(external_gates = ["rx"])

......

    783 gate_name: str = operation.name.name
    785 if gate_name not in self._custom_gates:
--> 786     raise_qasm3_error(
    787         f"Gate '{gate_name}' is not a custom gate and can not be marked as external.",
    788         span=operation.span,
    789     )
    791 gate_definition: qasm3_ast.QuantumGateDefinition = self._custom_gates[gate_name]
    792 op_qubits: list[qasm3_ast.IndexedIdentifier] = (
    793     self._get_op_bits(  # type: ignore [assignment]
    794         operation,
    795         self._global_qreg_size_map,
    796     )
    797 )

File /workspaces/pyqasm/pyqasm/exceptions.py:64, in raise_qasm3_error(message, err_type, span, raised_from)
     62 if raised_from:
     63     raise err_type(message) from raised_from
---> 64 raise err_type(message)

ValidationError: Gate 'rx' is not a custom gate and can not be marked as external.
```
The user should have the ability to specify any gate - user defined custom gates AND from the set of gates which we support, to be marked as external.

TheGupta2012: More specifically the gates which we decompose down to simpler gates like the [`u3`](https://github.com/qBraid/pyqasm/blob/3a7fa5c0636a7b8ac5033cd9ebc100becff142df/pyqasm/maps.py#L94), [`crx`](https://github.com/qBraid/pyqasm/blob/3a7fa5c0636a7b8ac5033cd9ebc100becff142df/pyqasm/maps.py#L327), etc. will benefit from this change.

pyqasm/visitor.py
```diff
@@ -759,6 +761,69 @@ def _visit_custom_gate_operation(
...
+            raise_qasm3_error(
+                f"Gate '{gate_name}' is not a custom gate and can not be marked as external.",
+                span=operation.span,
+            )
+
+        gate_definition: qasm3_ast.QuantumGateDefinition = self._custom_gates[gate_name]
```
TheGupta2012, line 791: If a basic gate is detected, we just offload the computation to the visit basic gate method then.

TheGupta2012:
Hey, thanks for the changes! I've given some comments on this and the linked PR. 

The only change I see is making the acceptable `external_gates` more generic and not confined to user defined custom gates. There may very well be a use case when someone wants to stop the decomposition of, say, a `crz` gate and pass that in the `external_gates` list. Since we support that gate natively in our unroller, it would not be possible to specify that in the `external_gates` variable with the current logic.

TheGupta2012:
Hey, @Sola85 just a heads up - we added some new commits to `main` so you'll probably need to rebase the PR against it

Sola85:
Hi @TheGupta2012, thanks for the feedback! I hadn't realized that pyqasm unrolls some gates into other gates and therefore thought it didn't make sense to mark non-custom gates as external. 

With my last commit, it should be possible to mark any gate as external (I've added a test that tests this for the `u3` gate). What is still missing is the unrolling of gates with multiple targets (e.g. `u3(0.5, 0.5, 0.5) q[0], q[1];`). Currently this doesn't work for external gates (and I think also not for custom gates). 
I was wondering if it made sense to offload this unrolling logic into somewhere outside of `_visit_*_gate_operation()`, since it is identical for all types of gates.

pyqasm/visitor.py
```diff
@@ -50,9 +50,10 @@ class QasmVisitor:
     Args:
         initialize_runtime (bool): If True, quantum runtime will be initialized. Defaults to True.
         record_output (bool): If True, output of the circuit will be recorded. Defaults to True.
+        external_gates (list[str]): List of custom gates that should not be unrolled.
```
TheGupta2012, line 53:
```suggestion
        external_gates (list[str]): List of gates that should not be unrolled.
```

TheGupta2012:
I agree, the unrolling of gates with multiple targets is not handled for custom operations. This is because of the assumption that any custom gate will eventually be broken down into simpler gates with either the user supplied definition or the one present in the `maps.py`.

We can have something like a `_broadcast_gate_operation` which will essentially use the [same logic present in `_visit_basic_gate_operation`](https://github.com/qBraid/pyqasm/blob/78000e80aa5844aee9cf99b2815bb549476ceab3/pyqasm/visitor.py#L657) for decomposing multiple targets but, as you suggested, make it more generic so that it can be used on every gate. This method would return a list of `QuantumGate` with the qubit arguments transformed according to the op count of gate.

A point to note is that this will only be required when we are dealing with external gates as any non-external gate will eventually be broken down into simpler gates already having this broadcast logic.

TheGupta2012:
Another thing to note is about the qubit depths. We recently added the support for calculating the `depth` of a quantum program in pyqasm and with this change, there would be some impact on the `depth` calculation logic.

Eg, if we are unrolling with an external gate `custom_1` - 

```c
OPENQASM 3.0;
qubit[3] q;
gate custom_1 a { 
     h a;
     x a;
} 
custom_1 q[0];
custom_1 q[1];
```
the final unrolled qasm would look like - 

```c
OPENQASM 3.0;
qubit[3] q;
custom_1 q[0];
custom_1 q[1];
```

Inside the `depth` calculation, we are [making a call to `unroll`](https://github.com/qBraid/pyqasm/blob/78000e80aa5844aee9cf99b2815bb549476ceab3/pyqasm/modules/base.py#L252) before calculating the depth using individual qubits. A couple of points to note here - 

1. Since the final `unrolled_qasm` only contains `custom_1` op, which might very well be the basis gate of a QC, the depth of the program should come out to be 1. But the depth we get is 2 as internally we visit the `custom_1` definition and update the qubit depths to 2 - 

```python
In [16]: import pyqasm

In [17]: qasm_str = """
    ...: OPENQASM 3.0;
    ...: qubit[3] q;
    ...: gate custom_1 a { 
    ...:      h a;
    ...:      x a;
    ...: } 
    ...: custom_1 q[0];
    ...: custom_1 q[1];"""

In [18]: mod = pyqasm.load(qasm_str)

In [19]: mod.unroll(external_gates = ["custom_1"])

In [20]: mod.unrolled_qasm.splitlines()
Out[20]: 
['OPENQASM 3.0;',
 'include "stdgates.inc";',
 'qubit[3] q;',
 'custom_1 q[0];',
 'custom_1 q[1];']

In [21]: mod.depth()
Out[21]: 2
```
This is incorrect as we want to get the depth of the current unrolled program and not unroll it yet again. 

2. The `_visit_external_gate_operation` should also update the qubit depths before returning the resulting gate. Moreover, when it is [calling the `_visit_basic_gate_operation`](https://github.com/qBraid/pyqasm/pull/59/files#diff-5b26d2cb7dbbf506dbaa06405eeb8e9b44a0786e2ac64ede3787b3e96a964207R797), qubit depths are implicitly getting updated which is not correct. 

This might expand the scope a bit and I think for now you can add a test in `tests/qasm3/test_depth.py` with a `@pytest.mark.skip` decorator highlighting this bug. Once this PR merges, I'll create a follow up issue to tackle it.

Sola85:
Changes since my last comment:
- I have isolated the broadcasting logic into a separate function which works for basic gates and external gates
- There is an additional test for broadcasting external gates
- I have added a test for depth-calculation of external gates, but this is currently skipped (and, as you said, could be addressed in a separate issue)

I have not yet extended the broadcasting logic to custom gates i.e. something like 
```python 
gate mygate() _gate_q_0,  {
  h _gate_q_0;
}
qubit[2] q;
mygate() q;
```
still does not work. But I think this is also not related to this PR and could be addressed separately

pyqasm/visitor.py
```diff
@@ -608,6 +611,71 @@ def _visit_gate_definition(self, definition: qasm3_ast.QuantumGateDefinition) ->
...
+                The length of this list indicates the number of time the gate is invoked.
+        Returns:
+            List of all executed gates.
+        """
+        result = []
+        for targets in all_targets:
+            result.extend(gate_function(*targets))
+        return result
+
+    def _update_qubit_depth(self, all_targets: list[list[qasm3_ast.IndexedIdentifier]]):
```
TheGupta2012, line 657: I think for clarity we should name it something like `_update_qubit_depths_for_gate` as we have a lot of functions where depths are being modified. Eventually we wanna move towards one method where we handle this updation logic for all basic operations like `measure`, `reset`, `gate` etc.

pyqasm/visitor.py
```diff
@@ -765,6 +810,67 @@ def _visit_custom_gate_operation(
...
+
+        if gate_name in self._custom_gates:
+            # Ignore result, this is just for validation
+            self._visit_custom_gate_operation(operation, inverse=inverse)
+            # Don't need to check if custom gate exists, since we just validated the call
+            gate_qubit_count = len(self._custom_gates[gate_name].qubits)
+        else:
+            # Ignore result, this is just for validation
+            self._visit_basic_gate_operation(operation, inverse=inverse)
+            # Don't need to check if custom gate exists, since we just validated the call
```
TheGupta2012, line 842:
```suggestion
            # Don't need to check if basic gate exists, since we just validated the call
```

pyqasm/visitor.py
```diff
@@ -765,6 +810,67 @@ def _visit_custom_gate_operation(
...
+            qasm3_ast.FloatLiteral(param) for param in self._get_op_parameters(operation)
+        ]
+
+        self._push_context(Context.GATE)
+
+        modifiers = []
+        if inverse:
+            modifiers = [qasm3_ast.QuantumGateModifier(qasm3_ast.GateModifierName.inv, None)]
+
+        def gate_function(*qubits):
```
TheGupta2012, line 855: Why is `op_parameters` not a parameter of the `gate_function`?

pyqasm/visitor.py
Sola85: Fixed.

pyqasm/visitor.py
Sola85: Fixed.

pyqasm/visitor.py
Sola85: The way it is currently implemented, `_broadcast_gate_operation()` expects a callable that only takes qubits as input. The intended use is that the non-qubit arguments get evaluated before calling `_broadcast_gate_operation`, e.g. using [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial), like [here](https://github.com/qBraid/pyqasm/pull/59/files#diff-5b26d2cb7dbbf506dbaa06405eeb8e9b44a0786e2ac64ede3787b3e96a964207R726). This way, `_broadcast_gate_operation` doesn't need to know about any non-qubit arguments.

But if you prefer the non-qubit arguments to be passed along explictly, I could also easily change that.

Sola85: For now, I have updated the docstring of `_broadcast_gate_operation` to make this point more clear.

TheGupta2012:
Thanks for working on this @Sola85 , the changes look good!

Sola85:
Thanks for the feedback and thanks for merging!

PR #83, opened 2024-11-16
linalg bug fix & cython integration

ryanhill1:
## Summary of changes

- Fixed small bug in `so_bidiagonalize`
- Migrated `kronecker_fator` function to Cython (~60% speedup)
- Migrated `so4_to_so2` function to Cython (~5% speedup)
- These initial Cython migrations weren't as much out of concern for performance, but rather done as more of an experiment and to learn how it works. Now that the structure is there, we may be able to migrate some of the other larger functions and get an actual performance improvement out of `kak_decomposition_angles`, if we wanted to.


src/pyqasm/linalg.py
```diff
@@ -42,16 +42,19 @@ def is_unitary(matrix: np.ndarray, rtol: float = 1e-5, atol: float = 1e-8) -> bo
     Returns:
         bool: True if the matrix is unitary, False otherwise.
     """
-    if matrix.shape[0] != matrix.shape[1]:
-        return False
+    try:
```
TheGupta2012, line 45: Can use `isinstance(matrix, np.ndarray)` for a tighter check.

TheGupta2012:
I like the idea here and the speed up is pretty good! Going forward, do you think it would be better to arrange the `.pyx` files in a `cython` folder and rename the `.pyx` file as the module name? Eg. - 
1. `src/pyqasm/linalg_cy.pyx` becomes `src/pyqasm/cython/linalg.pyx` and 
2.  `from pyqasm.linalg_cy` becomes `from pyqasm.cython.linalg import ...`

TheGupta2012 (approved):
Looks good! Thanks for the changes

PR #87, opened 2024-11-20
pyqasm validate via CLI

ryanhill1:
## Summary of changes

Added `pyqasm.cli` module with `typer` integration to enable using pyqasm as a command-line tool. For example:

```bash
$ pip install 'pyqasm[cli]'
$ pyqasm --help
Usage: pyqasm [OPTIONS] COMMAND [ARGS]...
$ pyqasm --version
pyqasm/0.1.0a1
$ pyqasm validate tests/cli/resources
tests/cli/resources/invalid1.qasm: error: Index 2 out of range for register of size 1 in qubit [validation]
Found errors in 1 file (checked 3 source files)
$ pyqasm validate tests/cli/resources --skip tests/cli/resources/invalid1.qasm
Success: no issues found in 2 source files
```

### DIFF
--- a/src/pyqasm/cli/validate.py
+++ b/src/pyqasm/cli/validate.py
@@ -13,6 +13,7 @@ Script to verify OpenQASM files
 
 """
 
+import logging
 import os
 from typing import Optional
 
@@ -20,6 +21,9 @@ import typer
 from rich.console import Console
 
 from pyqasm import load
+from pyqasm.exceptions import QasmParsingError, UnrollError, ValidationError
+
+logger = logging.getLogger(__name__)
 
 
 def validate_paths_exist(paths: Optional[list[str]]) -> Optional[list[str]]:
@@ -51,19 +55,17 @@ def validate_qasm(src_paths: list[str], skip_files: Optional[list[str]] = None) 
             return True
 
         skip_tag = "// pyqasm: ignore"
-        line_number = 0
 
         for line in content.splitlines():
-            line_number += 1
-            if 5 <= line_number <= 30 and skip_tag in line:
+            if skip_tag in line:
                 return True
-            if line_number > 30:
+            if "OPENQASM" in line:
                 break
 
         return False
 
     def validate_qasm_file(file_path: str) -> None:
-        with open(file_path, "r", encoding="ISO-8859-1") as f:
+        with open(file_path, "r", encoding="utf-8") as f:
             content = f.read()
 
         if should_skip(file_path, content):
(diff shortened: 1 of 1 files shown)


src/pyqasm/cli/validate.py
```diff
@@ -0,0 +1,123 @@
...
+            line_number += 1
+            if 5 <= line_number <= 30 and skip_tag in line:
+                return True
+            if line_number > 30:
+                break
+
+        return False
+
+    def validate_qasm_file(file_path: str) -> None:
+        with open(file_path, "r", encoding="ISO-8859-1") as f:
```
TheGupta2012, line 66: any particular reason for not choosing the `utf-8` encoding here?

src/pyqasm/cli/validate.py
```diff
@@ -0,0 +1,123 @@
...
+            content = f.read()
+
+        if should_skip(file_path, content):
+            return
+
+        try:
+            module = load(file_path)
+            module.validate()
+        except Exception as err:  # pylint: disable=broad-exception-caught
```
TheGupta2012, line 75: I think it is better to catch the `pyqasm.exceptions.ValidationError` exception here

src/pyqasm/cli/validate.py
```diff
@@ -0,0 +1,123 @@
...
+    def should_skip(filepath: str, content: str) -> bool:
+        if filepath in skip_files:
+            return True
+
+        skip_tag = "// pyqasm: ignore"
+        line_number = 0
+
+        for line in content.splitlines():
+            line_number += 1
+            if 5 <= line_number <= 30 and skip_tag in line:
```
TheGupta2012, line 58: How are `5` and `30` chosen here?

src/pyqasm/cli/validate.py
```diff
@@ -0,0 +1,123 @@
...
+                return True
+            if line_number > 30:
```
TheGupta2012, line 60: Here too

src/pyqasm/cli/validate.py
ryanhill1: No good reason, copied most of this code from the qbraid-cli headers check. But can definitely change it to `utf-8`. That would probably make more sense.

src/pyqasm/cli/validate.py
ryanhill1: Aren't there other unroll exceptions or parsing exceptions that are also possible? Want to make sure we catch all exceptions, cause if we don't the command will fail with error without providing any helpful info. Can definitely add more granularity in catching these exceptions though

src/pyqasm/cli/validate.py
ryanhill1: No good reason, carried over from qbraid-cli headers check. Basically defines the bounds of where we check for for the skip comment. But can be removed, or require that file skip comment is on the first line.

src/pyqasm/cli/validate.py
TheGupta2012: Yep, thanks for catching that

PR #88, opened 2024-11-25
Update build process with `cibuildwheel`

TheGupta2012:
## Summary of changes
This PR helps set up a template for any future integrations with `Cython` in our build pipelines. Have added platform and arch specific builds for each supported OS. Wheels are now built with `cibuildwheel` tool.

Reference for `yaml` structure and scripts : [Scikit Learn GitHub repo](https://github.com/scikit-learn/scikit-learn/tree/main/.github/workflows)

### CI/CD Workflow Updates:
* [`.github/workflows/main.yml`](diffhunk://#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3L5-R176): Added support for multiple OS and Python versions, and introduced a new `build_sdist` job for source distribution. Added a `code-coverage` job to handle test coverage outside of Docker containers.
* [`.github/workflows/release.yml`](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R9-R159): Added a `build_wheels` job similar to the main workflow, along with a `build_sdist` job. Updated the `pypi-publish` job to depend on these new jobs and download their artifacts.

### Test Suite Improvements:
* [`tests/cli/test_cli_commands.py`](diffhunk://#diff-cb3c9018cbc548519e82958f24b416e56eda72e67b5aba4fd3de3bd8a95961f8R17): Introduced a `normalize_output` function to standardize test output by stripping whitespace and replacing multiple spaces with a single space. Updated existing tests to use this function. [[1]](diffhunk://#diff-cb3c9018cbc548519e82958f24b416e56eda72e67b5aba4fd3de3bd8a95961f8R17) [[2]](diffhunk://#diff-cb3c9018cbc548519e82958f24b416e56eda72e67b5aba4fd3de3bd8a95961f8R42-R46) [[3]](diffhunk://#diff-cb3c9018cbc548519e82958f24b416e56eda72e67b5aba4fd3de3bd8a95961f8L50-R56) [[4]](diffhunk://#diff-cb3c9018cbc548519e82958f24b416e56eda72e67b5aba4fd3de3bd8a95961f8L76-R82)
* [`tests/qasm3/test_entrypoint.py`](diffhunk://#diff-152f790c76967d54b981214cc05a4e9b72ff19730bd16dac5a32810b76e89c06R23-R35): Consolidated resource paths by defining `QASM_RESOURCES_DIR` and updating file paths accordingly.


ryanhill1 (commented):
Why is it required we move to minimum Python 3.11? From what I found `cibuildwheel` supports Python 3.10: https://cibuildwheel.pypa.io/en/stable/

Right now qBraid-SDK supports Python 3.9, so incorporating PyQASM as a core dependency we're already having to bump SDK to minimum Python 3.10. This is ok because Python 3.9 EOL is October 31, 2025, which isn't too far away, so justifiable. But Python 3.10 EOL isn't until October 31, 2026. So I think we should definitely still provide support for Python 3.10.

Think about not just the qBraid-SDK, but other Python packages in the open-source ecosystem as well. If any of those packages want to use PyQASM as a dependency, they'll also have to bump their minimum Python requirement to match ours. Which could be a deal breaker for many.

TheGupta2012:
> Why is it required we move to minimum Python 3.11? From what I found `cibuildwheel` supports Python 3.10: https://cibuildwheel.pypa.io/en/stable/
> 
> Right now qBraid-SDK supports Python 3.9, so incorporating PyQASM as a core dependency we're already having to bump SDK to minimum Python 3.10. This is ok because Python 3.9 EOL is October 31, 2025, which isn't too far away, so justifiable. But Python 3.10 EOL isn't until October 31, 2026. So I think we should definitely still provide support for Python 3.10.
> 
> Think about not just the qBraid-SDK, but other Python packages in the open-source ecosystem as well. If any of those packages want to use PyQASM as a dependency, they'll also have to bump their minimum Python requirement to match ours. Which could be a deal breaker for many.

Thanks pointing this out @ryanhill1 . I was wrongly thinking that the `cibuildwheel` version which we use will only support "builds" for Python 3.11 and greater. But, turns out, this python version is needed only for "running" `cibuildwheel` command and unrelated to the supported Python package versions.

The project specific wheels can still be built for any lower version supported by `cibuildwheel` (as you mentioned in the comment)

bin/test_sdist.sh
ryanhill1, line 1: Same comment as above. Would prefer not to include `cd` commands. Can instead use temporary directory for virtual environment. And should optionally allow specifying full path to `pyqasm` directory so that the script isn't restricted to being called only from the source code directory.

Also, it would be preferable if we didn't have to explicitly list out the optional dependencies in this script needed to run tests. Instead of installing the `.tar.gz`, we could use the [install_wheel_extras.sh](https://github.com/qBraid/pyqasm/blob/main/bin/install_wheel_extras.sh) script to install the distribution and the CLI extras at the same time. To do this without a `cd`, you would also probably need to pass the path to the script as argument e.g.

```bash
$BIN_PATH/install_wheel_extras.sh $PYQASM_PATH/dist --extra cli
```

Otherwise, every time we update the CLI dependencies, we will also have to update them in this script. Easier to maintain `pyproject.toml` as the one source of truth if we can, and trigger all builds & tests from what's specified in there.

tests/qasm3/test_entrypoint.py
```diff
@@ -20,17 +20,19 @@
 from pyqasm.exceptions import ValidationError
 from tests.utils import check_unrolled_qasm
 
+QASM_RESOURCES_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "resources/qasm")
```
ryanhill1, line 23:
```suggestion
QASM_RESOURCES_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "resources", "qasm")
```

ryanhill1 (changes requested):

.github/workflows/release.yml
```diff
@@ -6,24 +6,157 @@ on:
...
         with:
-          python-version: '3.12'
-
-      - name: Build binary wheel + source tarball
+          python-version: '3.11'
         run: |
-          python3 -m pip install --upgrade pip build
-          python3 -m build
+          python -m pip install --upgrade pip
+          python -m pip install twine
```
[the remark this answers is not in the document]
TheGupta2012: I thought that the publish package to PyPI might need the `twine` package pre-installed but they must have it as a dependency in the action

bin/test_sdist.sh
TheGupta2012: Yes, this is a cleaner way to approach the installation and tests, thanks! 
Building on similar lines, I also found an option [`CIBW_TEST_EXTRAS`](https://cibuildwheel.pypa.io/en/stable/options/#test-extras) which installs the wheels with our specified extra flags. This eliminates the need for specifying these deps in `bin/cibw/pre_test.sh`

bin/cibw/pre_test.sh
ryanhill1, line 1: Could you explain a little bit what the purpose of this script is? From what I can tell, in the "Build and Test Wheels" step, you are setting environment variable `CIBW_BEFORE_TEST` equal to the output of this script. Is that correct? If so, why is that useful or necessary? Don't we already have the platform info for each build from the `os` matrix which contains each `platform_id`?

bin/install_wheel_extras.sh
ryanhill1, line 1: Nice

bin/test_sdist.sh
ryanhill1, line 1: Looks great

bin/cibw/pre_test.sh
TheGupta2012: I think the main motive was to add the test deps (which are not needed anymore) and platform specific info was added on top of it. Will remove this.

> P.S. the value of `CIBW_BEFORE_TEST` variable is not set to the output of the script, but rather the value of the var is a command executed by `cibuildwheel` before each test is run for the package wheel.

TheGupta2012:
@ryanhill1 I think all changes are done, thanks a ton for the review!

ryanhill1 (approved):

PR #112, opened 2024-12-12
multi-bit equality branching

### LINKED ISSUE #52: [FEATURE] Decompose branching statement containing `bit` register
### Feature Description

`openqasm3` supports branching statements of the form - 
```c
...
qubit[3] q; 
h q; 
bit[3] c; 
c = measure q; 
if (c == 4) { ... }
``` 
where `c` is a 3 bit classical register. Currently, `pyqasm` only supports unrolling [single bit measurement results](https://github.com/qBraid/pyqasm/blob/main/tests/qasm3/test_if.py#L57) but we aim to support unrolling of this shorthand notation. 

The motivation for this feature is to enable program execution for QC vendors which only supports single bit level branches. Using the above, a generic register comparison can be unrolled to "cascaded" single bit branching statements and, thus, run on the system. 

### Implementation (Optional)

- The comparison can be divided into 3 broad categories - `==`, `>=` and `<=`. 
- Register comparisons will be something like `if (c == 6) { // execute this block of statements}`.
- Whenever a "full register" is encountered in the branching statement we do the following - 
    - Get the size of the register ,say, `4`
    - Represent the RHS i.e. `6` in `4` bits - `0110`
    - For each bit of the register, compare the corresponding bit of the RHS - 
   
        ```c
           if (c[0] == 0) 
           {  if (c[1] == 1) 
              {  if (c[2] == 1) 
                 {  if (c[3] == 0) 
                      { // execute this block } 
                  }
               }
            }
          ```
    - `==` can be replaced with `>=` or `<=` for the respective binary operator 
- Major chunk of processing would be done in [`_visit_branching_statement` method](https://github.com/qBraid/pyqasm/blob/d77251c24e9f744a40ab6a5363dbb891c976c850/pyqasm/visitor.py#L1071) of the visitor module. 
- Essentially, we just construct a new `if_block` for the branching statement which will contain these cascading if conditions. The last `if_block` will contain the block of statements

arulandu:
## Summary of changes
Unroll branching for multi-bit equality. Closes #52.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ In [4]: from pyqasm import dumps
 In [5]: dumps(module).splitlines()
 Out[5]: ['OPENQASM 3.0;', 'qubit[2] q;', 'h q;']
 ```
-- Added support for unrolling multi-bit equality branching.
+- Added support for unrolling multi-bit branching with `==`, `>=`, `<=`, `>`, and `<`.
 
 ### Improved / Modified
  - Refactored the initialization of `QasmModule` to remove default include statements. Only user supplied include statements are now added to the generated QASM code ([#86](https://github.com/qBraid/pyqasm/pull/86))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,37 @@ In [4]: from pyqasm import dumps
 In [5]: dumps(module).splitlines()
 Out[5]: ['OPENQASM 3.0;', 'qubit[2] q;', 'h q;']
 ```
-- Added support for unrolling multi-bit branching with `==`, `>=`, `<=`, `>`, and `<`.
+- Added support for unrolling multi-bit branching with `==`, `>=`, `<=`, `>`, and `<`. Usage example -
+```python
+In [1]: from pyqasm import loads
+
+In [2]: module = loads(
+   ...: """OPENQASM 3.0;
+   ...: include "stdgates.inc";
+   ...: qubit[1] q;
+   ...: bit[4] c;
+   ...: if(c == 3){
+   ...:     h q[0];
+   ...: }
+   ...: """)
+
+In [3]: module.unroll()
+
+In [4]: dumps(module)
+OPENQASM 3.0;
+include "stdgates.inc";
+qubit[1] q;
+bit[4] c;
+if (c[0] == false) {
+  if (c[1] == false) {
+    if (c[2] == true) {
+      if (c[3] == true) {
+        h q[0];
+      }
+    }
+  }
+}
+```
 
 ### Improved / Modified
  - Refactored the initialization of `QasmModule` to remove default include statements. Only user supplied include statements are now added to the generated QASM code ([#86](https://github.com/qBraid/pyqasm/pull/86))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ In [4]: from pyqasm import dumps
 In [5]: dumps(module).splitlines()
 Out[5]: ['OPENQASM 3.0;', 'qubit[2] q;', 'h q;']
 ```
+- Added support for unrolling multi-bit equality branching.
 
 ### Improved / Modified
  - Refactored the initialization of `QasmModule` to remove default include statements. Only user supplied include statements are now added to the generated QASM code ([#86](https://github.com/qBraid/pyqasm/pull/86))
--- a/src/pyqasm/transformer.py
+++ b/src/pyqasm/transformer.py
@@ -228,7 +228,12 @@ class Qasm3Transformer:
     @staticmethod
     def get_branch_params(
         condition: Expression,
-    ) -> tuple[Optional[int], str, Optional[Union[BinaryOperator, UnaryOperator]], Optional[bool]]:
+    ) -> tuple[
+        Optional[int],
+        str,
+        Optional[Union[BinaryOperator, UnaryOperator]],
+        Optional[Union[bool, int]],
+    ]:
         """
         Get the branch parameters from the branching condition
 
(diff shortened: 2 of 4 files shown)


src/pyqasm/transformer.py
```diff
@@ -225,15 +226,24 @@ def transform_gate_params(
             gate_op.argument = Qasm3Transformer.transform_expression(gate_op.argument, param_map)
 
     @staticmethod
-    def get_branch_params(condition: Any) -> tuple[Optional[int], str, Optional[bool]]:
+    def get_branch_params(
+        condition: Expression,
+    ) -> tuple[
```
TheGupta2012, line 231: It is not immediately evident what this function returns until we gloss through the docs. I think defining a `namedtuple` would make the definition clearer, something like - 
```python
from collections import namedtuple 
BranchParams = namedtuple("BranchParams", ["reg_id", "reg_name", "op", "rhs_val"])
```
This `BranchParams` can be used as the return type of the `get_branch_params` function. If the values are getting updated in visitor, feel free to used `dataclass`.

tests/qasm3/test_if.py
```diff
@@ -198,7 +267,7 @@ def test_incorrect_if():
            h q;
            measure q->c;
 
-           if(c[0] >= 1){
+           if(c[0] >> 1){
```
TheGupta2012, line 270: Is this a typo?

src/pyqasm/visitor.py
```diff
@@ -1331,44 +1331,83 @@ def _visit_branching_statement(
...
 
+                if op == qasm3_ast.BinaryOperator[">"]:
+                    op = qasm3_ast.BinaryOperator[">="]
+                    rhs_value += 1
+                elif op == qasm3_ast.BinaryOperator["<"]:
+                    op = qasm3_ast.BinaryOperator["<="]
+                    rhs_value -= 1
+
+                size = self._global_creg_size_map[reg_name]
+                rhs_value_str = bin(int(rhs_value))[2:].zfill(size)
```
TheGupta2012, line 1384: Can you double check if this works for negative values as well? Will be nice to have a test for negative values

tests/qasm3/test_if.py
```diff
@@ -136,6 +129,79 @@ def test_complex_if():
...
+        h q[0];
+    } else {
+        x q[0];
+    }
+    """
+    expected_qasm = """OPENQASM 3.0;
+    include "stdgates.inc";
+    qubit[1] q;
+    bit[4] c;
+    if (c[3] == false) {
```
TheGupta2012, line 160: I think using the big-endian notation will be more natural here because we mostly define a bit register like `bit[4] c = "0011";`. In this case, `c[0]` will be `false`, `c[1]` as `false`, so on.

TheGupta2012:
```python
In [6]: bin(18)[2:].zfill(8)
Out[6]: '00010010'
```
In Python too, the representation is big-endian

TheGupta2012:
Thanks for the work @arulandu ! Suggested some changes but overall looks good.

tests/qasm3/test_if.py
arulandu: I supported >= now? Maybe I'm misreading this. Shouldn't this pass?

src/pyqasm/visitor.py
```diff
@@ -1331,44 +1331,83 @@ def _visit_branching_statement(
...
+                else_block = self.visit_basic_block(statement.else_block)
+
+                def ravel(i):
```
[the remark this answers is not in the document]
arulandu: Fair enough. Will do

src/pyqasm/visitor.py
```diff
@@ -1331,44 +1331,83 @@ def _visit_branching_statement(
...
+                    r = rhs_value_str[i] == "1"
+                    if (op == qasm3_ast.BinaryOperator[">="] and not r) or (
+                        op == qasm3_ast.BinaryOperator["<="] and r
+                    ):
+                        # ith-bit doesn't affect condition -> skip
```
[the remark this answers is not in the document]
arulandu: Yeah, it's def a "think about it" comment, but I can give an example

tests/qasm3/test_if.py
arulandu: Got it, will fix

src/pyqasm/visitor.py
arulandu: How can we support negatives at all if LHS is bit[]?

TheGupta2012: Well negatives can be implemented using something like a sign bit or the two's complement method. So we can compare against that representation for any negative number. 

Will have to see the repr in Python and accordingly deal with it. Note that this reserves 1 bit for the sign.

tests/qasm3/test_if.py
TheGupta2012: Ah gotcha, my bad. Didn't see that this was testing an incorrect if.

src/pyqasm/visitor.py
arulandu: Yes, but do we want to restrict the representation like this? This means that bit[4] encodes -7 to 8 as opposed to 0 to 15. Adding this would make users have to rememeber to keep the sign bit, which doesn't seem conventional. I feel like most quantum algorithms with classical measurement record unsigned values?

TheGupta2012: Yea that's a valid point, let's keep the logic for non-negative numbers only

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ In [4]: from pyqasm import dumps
 In [5]: dumps(module).splitlines()
 Out[5]: ['OPENQASM 3.0;', 'qubit[2] q;', 'h q;']
 ```
+- Added support for unrolling multi-bit branching with `==`, `>=`, `<=`, `>`, and `<`.
```
TheGupta2012, line 41: Can you add an example here showing the qasm before and after unrolling a program?

src/pyqasm/transformer.py
```diff
@@ -47,6 +48,16 @@
 
 # mypy: disable-error-code="attr-defined, union-attr"
 
+BranchParams = NamedTuple(
```
TheGupta2012, line 51: nice!

Issue #89, opened 2024-11-26
[DOCS] Explicitly list `**kwargs` in `QasmModule.unroll` function arg / docstring


ryanhill1:
The `QasmModule.unroll()` method currently just specifies `**kwargs` in it's function argument:

https://github.com/qBraid/pyqasm/blob/5a0c7b352b78d7557f0c8ed5eba5595d604677ff/src/pyqasm/modules/base.py#L487

On the surface, this seems like a convenient short-hand because these `kwargs` will be passed directly to the `QasmVisitor.__init__` method. However, from the [API reference docs page](https://sdk.qbraid.com/projects/pyqasm/en/latest/stubs/pyqasm.QasmModule.html#pyqasm.QasmModule), this isn't clear: 

![Screenshot 2024-11-26 at 9 29 18 AM](https://github.com/user-attachments/assets/d6f58aec-8f38-405c-9857-e11f2bd22714)

Since the `unroll` method is public facing, it would be good to document the available kwargs explicitly so that users don't have to dig through the source code to track the kwargs through to initialization of `QasmVisitor`.

micpap25:
Two thoughts @ryanhill1; This has been accomplished and can be closed as all the `**kwargs` are listed in the docstring. But, this could be documented further using a `TypedDict` perhaps?

PR #122, opened 2025-01-11
QasmModule Circuit Drawer

### LINKED ISSUE #61: [FEATURE] Add `QasmModule.draw()` method
We would like to add a `QasmModule.draw()` method that generates a circuit diagram from a parsed OpenQASM program (i.e. an `openqasm3.ast.Program` object). This method can either generate either a raw ASCII string or leverage external visualization tools like [pylatex](https://github.com/JelteF/PyLaTeX), but should *not* depend on any other quantum packages outside of the `openqasm3`.

Because of the large number of gates and operations supported by OpenQASM, we will restrict this first implementation to a smaller gate-set to make it a little more approachable. For this, you can use the following code to generate random qasm strings to use as the test data/files for the circuit drawer. Once the circuit drawer is reliably working for any circuits generated by this function, then we will consider this issue closed, and open a new issue that expands the scope.

```bash
pip install qbraid
pip install qiskit
pip install qiskit-qasm3-import --no-deps
```

```python
import random

from pyqasm import load

from qbraid import random_circuit, transpile

qiskit_circuit = random_circuit("qiskit", measure=random.choice([True, False]))

qasm_str = transpile(qiskit_circuit, random.choice(["qasm2", "qasm3"]))

pyqasm_module = load(qasm_str)

pyqasm_module.draw()
```

arulandu:
## Summary of changes
Added circuit drawing functionality for QasmModule. Closes #61.


pyproject.toml
```diff
@@ -32,7 +32,7 @@ classifiers = [
     "Operating System :: Unix",
     "Operating System :: MacOS",
 ]
-dependencies = ["numpy", "openqasm3[parser]>=1.0.0,<2.0.0"]
+dependencies = ["numpy", "matplotlib", "openqasm3[parser]>=1.0.0,<2.0.0"]
```
ryanhill1, line 35: let's make matplotlib an optional dependency under a `visualization` extra  instead of a core dependency e.g.

```bash
pip install 'pyqasm[visualization]'
```

See: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements

and then within the `.draw()` method we can catch any `ImportError` and re-raise with an exception that prompts them to install the visualization extra. For example:

https://github.com/Qiskit/rustworkx/blob/752555d2b22000784464db90187c3afba60f72b1/rustworkx/visualization/matplotlib.py#L198-L205

tests/qasm3/test_printer.py
ryanhill1, line 1: Does this currently work with custom gates as well? If so, can you write a test case for that?

ryanhill1 (changes requested):

tests/qasm3/test_printer.py
arulandu: yes they work

ryanhill1 (changes requested):
This progress looks good! A couple things:

1. **Gate angle formatting**: For parameterized gates, the angle below the gate's name goes all the way out to the edges of the gate box, and for negative numbers, it bleeds over the sides. I think remove the parentheses around the angle and making the font size of the angle slightly smaller than the gate name could make it fit better, and allow the main focus to be on the gate name itself.

e31686c4-ff6b-409d-88f3-7bcb4fb99731



2. **Support for barriers**: Add support for barriers. Currently being omitted from the diagram:

```python
from pyqasm import loads

qasm3 = """
OPENQASM 3;
include "stdgates.inc";
qubit[2] q;
h q;
barrier q[0], q[1];
cnot q[0], q[1];
"""

module = loads(qasm3)

module.draw()
```

3f038bb3-789f-42e2-a7ee-35f5348cfe8b

3. **Measurement gate formatting**:
- Measurement gates are currently being stacked on top of one another which makes it difficult to read. Would be better to space them out.
- The arrow pointing to the classical bit to measure currently goes over that bottom line.
- Let's make the arrow for the measure a little bit smaller
- The current color for the measurement gates is a bit dark. Maybe a slightly lighter grey?

Ours:

```python
from pyqasm import loads

qasm3 = """
OPENQASM 3;
include "stdgates.inc";
qubit[2] q;
bit[2] b;
h q;
cnot q[0], q[1];
b = measure q;
"""

module = loads(qasm3)

module.draw()
```

ed87f7ca-2532-4a62-b66a-3e59a051c6d0-1

vs. qiskit:

```python
from qbraid import transpile

qiskit_circuit = transpile(qasm3, "qiskit")

qiskit_circuit.draw(output='mpl')
```

849cf702-7fd0-4246-9880-10235934858b

*Note*: I'm not saying to copy the way they've done it, but we can both it agree that theirs looks very clean. So we can perhaps take inspiration from the elements which make their diagrams visually appealing.

4. **Optional to exclude idle wires**: Can you add an option `idle_wires` of type `bool` which defaults to `True`, but if `False`, excludes any qubits and clbits that don't have any operation on them? For example:

```python
from pyqasm import loads

qasm3 = """
OPENQASM 3;
include "stdgates.inc";
qubit[2] q;
bit[2] b;
rx(-pi/2) q[0];
"""

module = loads(qasm3)

module.draw()
```
fa225092-cfeb-4ccc-a85f-bf66cb67b276

```python
module.draw(ide_wires=False)
```

e31686c4-ff6b-409d-88f3-7bcb4fb99731

5. **Duplicate image from draw output**: For some reason, when a call the `.draw()` method within my VS code notebook, it is "drawing" it twice. See images below. Is the same thing happening for you? See below:

Screenshot 2025-01-15 at 9 15 59 AM

6. **Split lines for long diagrams**: For circuit diagrams that are very long, we should at a certain point, split it onto two lines so that it is readable. Currently the font just gets smaller and smaller into oblivion

src/pyqasm/modules/qasm3.py
```diff
@@ -48,3 +49,7 @@ def accept(self, visitor):
         final_stmt_list = visitor.finalize(unrolled_stmt_list)
 
         self._unrolled_ast.statements = final_stmt_list
+
+    def draw(self, **kwargs):
```
ryanhill1, line 53: Can you explicitly list out the available kwargs here and notate them in the docstring? With just kwargs the function usage is cryptic on docs page and forces people to dig into source code. Need to fix this for other methods as well e.g. https://github.com/qBraid/pyqasm/issues/89

ryanhill1 (commented):
Fix merge conflicts

arulandu:
TODO: support QuantumPhaseStatement

ryanhill1:
I think qbraid-cli might have a bug with the headers formatting, try setting `qbraid-cli==0.9.4` in `[testenv:headers]` in `tox.ini` and that should fix the header bugs + the file that looks like it has 2,000+ lines changed

TheGupta2012:
Amazing work @arulandu ! Here are some minor comments - 

1. **Add qubit labels for multiple lines** : For circuits extending to multiple lines, it would be nice to have qubit markers for the wires. They are currently being omitted whenever the program extends - 
```python
import pyqasm
test = """
OPENQASM 3.0;
include "stdgates.inc";
qubit[2] q;
gate custom a, b {
    h a;
    z a;
    z a;
    y a;
    x a;
    rx(0.5) a;
}
for int i in [0:2] {
    custom q;
}
"""
circ = pyqasm.loads(test)
circ.draw()
```
image

2. **External gates should be drawn as is** : Given that we provide support for external gates, we should not unroll them before drawing - 

```python 
import pyqasm
test = """
OPENQASM 3.0;
include "stdgates.inc";
qubit[2] q;
gate custom a, b {
    h a;
    z a;
    z a;
    y a;
    x a;
    rx(-0.5) a;
}
custom q;

"""
circ = pyqasm.loads(test)
circ.unroll(external_gates = ["custom"])
circ.draw()
```

image

3. **Support for reset** : I believe reset ops are not supported right now. I guess you can open an issue for extending the drawer with resets or add them in this PR, whichever works

TheGupta2012:
> I think `qbraid-cli` might have a bug with the headers formatting, try setting `qbraid-cli==0.9.4` in `[testenv:headers]` in `tox.ini` and run `tox -e linters` again and that should fix the header bugs + the file that looks like it has 2,000+ lines changed. If it doesn't, let me know.
> 
> Also, what's causing all of the different CI build workflows to fail?

@arulandu I think same applies for the `ctrl` modifier PR where we're seeing a very big change in the `src/pyqasm/maps/gates.py`

TheGupta2012:
@arulandu I think you need to add `qbraid` in the [`test` extra for `pyqasm`](https://github.com/qBraid/pyqasm/pull/122/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R44). This is why the [tests are failing in the CI pipeline](https://github.com/qBraid/pyqasm/actions/runs/12873997411/job/35892589317?pr=122#step:4:600).

arulandu:
@TheGupta2012 
I omitted labels on 1. since this is what qiskit does. I can add them back them though. For external gates, until draw supports logical constructs like while/if/etc, we have to unroll inside the draw. I could support `.draw(external=...)`, but in my opinion, it makes more sense to support this when our drawer does support logical constructs so that we don't need to unroll inside the draw. This feels like a separate issue on its own. For 3, I can open an issue and get to it soon

tests/test_mpl_draw.py
```diff
@@ -0,0 +1,94 @@
...
+
+import pytest
+
+from pyqasm.entrypoint import loads
+from pyqasm.printer import mpl_draw
+
+pytest.importorskip("matplotlib", reason="Matplotlib not installed.")
+
+
+def _check_fig(_, fig):
```
TheGupta2012, line 23: `_check_fig` functionality should check for the expected elements rather than just asserting that the `ax.texts` is present. The correct way to do the testing should be to compare the images generated by the drawer. 

Potentially we can use `matplotlib.testing.compare.compare_images` method. We build an `expected` imgs dir which contains the expected plots, read them for each test and compare against the generated image.

src/pyqasm/printer.py
```diff
@@ -0,0 +1,590 @@
...
+            raise ValueError(f"Unsupported statement: {statement}")
+        if isinstance(statement, ast.QuantumGate):
+            qubits = [_identifier_to_key(q) for q in statement.qubits]
+            depth = 1 + max(depths[q] for q in qubits)
+            for q in qubits:
+                depths[q] = depth
+        elif isinstance(statement, ast.QuantumMeasurementStatement):
+            qubit_key = _identifier_to_key(statement.measure.qubit)
+            if statement.target is None:
+                raise NotImplementedError("Stand-alone measurement statements not yet supported.")
```
ryanhill1, line 193: Now that https://github.com/qBraid/pyqasm/pull/141 is merged can you also implement this?

src/pyqasm/printer.py
```diff
@@ -0,0 +1,590 @@
...
+    statement: QuantumStatement, line_nums: dict[tuple[str, int], int], ax: plt.Axes, x: float
+):
+    if isinstance(statement, ast.QuantumGate):
+        args = [Qasm3ExprEvaluator.evaluate_expression(arg)[0] for arg in statement.arguments]
+        lines = [line_nums[_identifier_to_key(q)] for q in statement.qubits]
+        _mpl_draw_gate(statement, args, lines, ax, x)
+    elif isinstance(statement, ast.QuantumMeasurementStatement):
+        qubit_key = _identifier_to_key(statement.measure.qubit)
+        if statement.target is None:
+            raise NotImplementedError("Stand-alone measurement statements not yet supported.")
```
ryanhill1, line 425: And this one?

ryanhill1 (changes requested):

tests/test_mpl_draw.py
arulandu: wouldn't we need to regenerate the test img dir every time we change styling or smthn?

ryanhill1 (changes requested):
@arulandu I added a couple basic tests to compare against images for you that should be sufficient for now.

I did find one bug though with how barriers are displayed when moments don't line up:

```python
from pyqasm import draw

qasm3 = """
OPENQASM 3;
include "stdgates.inc";
qubit[2] q;
x q[0];
barrier q;
"""

draw(qasm3, filename="circuit.png")
```

circuit-1

When applied across all qubits, the barrier should form a single vertical line, not be broken across qubits.

TheGupta2012:
> @arulandu I added a couple basic tests to compare against images for you that should be sufficient for now.
> 
> I did find one bug though with how barriers are displayed when moments don't line up:
> 
> ```python
> from pyqasm import draw
> 
> qasm3 = """
> OPENQASM 3;
> include "stdgates.inc";
> qubit[2] q;
> x q[0];
> barrier q;
> """
> 
> draw(qasm3)
> ```
> 
> circuit-1
> 
> When applied across all qubits, the barrier should form a single vertical line, not be broken across qubits.

I think this is because of how [`barriers are added in a moment`](https://github.com/qBraid/pyqasm/pull/122/files#diff-5b98d0217c1aad1d0cac2c01c64abdc7b640654cba8734b2299e4a92ec539519R222). I feel this is expected `draw` behavior since we are [`unrolling the program before the draw`](https://github.com/qBraid/pyqasm/pull/122/files#diff-5b98d0217c1aad1d0cac2c01c64abdc7b640654cba8734b2299e4a92ec539519R113). This will break down each barrier to single qubits and thus add each barrier statement to distinct moments. - 

### Possible solution
Add something like `unroll_barriers = False` param in `unroll` which will only do semantic check and leave them as is. Can be useful for `draw` so that we can find the qubits on which a barrier is applied. Once we have that, we can identify the qubit which has the max depth at the barrier application point and add barrier to that moment. Eg - 

image

In above image , we will apply 2 barriers - 
1. Barrier 1 - q2[0] -> q2[3] , q2[1] will have max depth (2) at the barrier application, so barrier is added to moment at depth 2
2. Barrier 2 - q[0] -> q[1], q[0] will again have max depth (2) at the barrier application, so barrier is added to moment at depth 2

ryanhill1 (approved):

PR #133, opened 2025-02-13
created solovay kitaev algorithm

AshwinKaliyaperumal:

### DIFF
--- a/src/solovay_kitaev.py
+++ b/src/solovay_kitaev.py
@@ -1,0 +1,158 @@
+import numpy as np
+from typing import List, Tuple
+
+class SU2Matrix:
+    """Class representing a 2x2 Special Unitary matrix."""
+    def __init__(self, matrix: np.ndarray):
+        self.matrix = matrix
+    
+    def __mul__(self, other: 'SU2Matrix') -> 'SU2Matrix':
+        return SU2Matrix(np.dot(self.matrix, other.matrix))
+    
+    def dagger(self) -> 'SU2Matrix':
+        """Returns the conjugate transpose."""
+        return SU2Matrix(self.matrix.conj().T)
+    
+    def distance(self, other: 'SU2Matrix') -> float:
+        """Calculates the operator norm distance between two matrices."""
+        diff = self.matrix - other.matrix
+        return np.linalg.norm(diff)
+
+def group_commutator(a: SU2Matrix, b: SU2Matrix) -> SU2Matrix:
+    """Compute the group commutator [a,b] = aba^{-1}b^{-1}."""
+    return a * b * a.dagger() * b.dagger()
+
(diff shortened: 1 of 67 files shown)


src/solovay_kitaev.py
```diff
@@ -0,0 +1,202 @@
...
+    best_dist = float('inf')
+    best_gate = None
+    
+    for gate in basic_gates:
+        dist = target.distance(gate)
+        if dist < best_dist:
+            best_dist = dist
+            best_gate = gate
+    
+    return best_gate
```
TheGupta2012, line 45: The approximating sequence length should not be 1 and is set by us. The user might have a requirement for more depth, but we will set this initially as <= 16. Ref - [https://arxiv.org/pdf/quant-ph/0505030](https://arxiv.org/pdf/quant-ph/0505030) 

image

##

image

I also believe @PranavTupe2000's code could be beneficial here, as discussed in our weekly meeting

src/solovay_kitaev.py
```diff
@@ -0,0 +1,202 @@
...
+
+        v_approx = v_sequence[0]
+        for gate in v_sequence[1:]:
+            v_approx = v_approx * gate
+        
+        w_approx = w_sequence[0]
+        for gate in w_sequence[1:]:
+            w_approx = w_approx * gate
+
+        result_sequence.extend([best_v, best_w, v_approx.dagger(), w_approx.dagger()])
```
TheGupta2012, line 109: I believe the `result_sequence` variable should also contain the $\mathbf{U}_{n-1}$ term (Ref: [Page 5, Line 3](https://arxiv.org/pdf/quant-ph/0505030)). Is there a reason to omit that here because of the recursive function structure?

src/solovay_kitaev.py
```diff
@@ -0,0 +1,202 @@
...
+    
+    # Calculate final error
+    final_approx = result_sequence[0]
+    for gate in result_sequence[1:]:
+        final_approx = final_approx * gate
+    final_error = target.distance(final_approx)
+    
+    return result_sequence, final_error
+
+def solovay_kitaev(target: np.ndarray, basic_gates: List[np.ndarray], depth: int = 3) -> List[np.ndarray]:
```
TheGupta2012, line 119: For a smoother interface, I believe `basis_gates` parameter should accept the [`BasisSet` enum](https://github.com/qBraid/pyqasm/blob/ef34299e3326ca5ef1d2bf24f101f17a87b6bb96/src/pyqasm/elements.py#L103) as a param and internally derive the basis gates from the [`BASIS_GATE_MAP`](https://github.com/qBraid/pyqasm/blob/ef34299e3326ca5ef1d2bf24f101f17a87b6bb96/src/pyqasm/maps/gates.py#L1090)

TheGupta2012 (changes requested):
Hi @AshwinKaliyaperumal , thanks for working on this!

PR #141, opened 2025-02-15
Support standalone measurement

### LINKED ISSUE #139: Assertion error on valid-standalone measurement
In OpenQASM grammar, it seems that it might not be required that measurement statements make reference to a classical (target) qubit.

The [openqasm3.ast.QuantumMeasurementStatement](https://github.com/openqasm/openqasm/blob/bb923eb9a84fdffe1ba6fc3c20d0b47a131523d9/source/openqasm/openqasm3/ast.py#L582-L590) is documented as:

> Stand-alone statement of a quantum measurement, *potentially* assigning the result to a classical variable.

and specifies field `target` of type `Optional[Union[IndexedIdentifier, Identifier]]`.

This makes me think that specifying a target (classical) bit might not be semantically necessary.

If this is true, the following should be valid:

```openqasm
OPENQASM 3.0;
include "stdgates.inc";

qubit[3] q;
bit[3] b;

h q[0];
cx q[0], q[1];
measure q;
```

However, in `pyqasm.unroll()`, it raises an `AssertionError`:

https://github.com/qBraid/pyqasm/blob/0fd2956c729c637df6669a7d4a83e3d6725e36b8/src/pyqasm/visitor.py#L441-L443

arulandu:
## Summary of changes
Support standalone measurements that do not store result in a classical register. Closes #139.

### DIFF
--- a/src/pyqasm/visitor.py
+++ b/src/pyqasm/visitor.py
@@ -343,8 +343,11 @@ class QasmVisitor:
         original_size_map = reg_size_map
 
         if isinstance(operation, qasm3_ast.QuantumMeasurementStatement):
-            assert operation.target is not None
-            bit_list = [operation.measure.qubit] if qubits else [operation.target]
+            if qubits:
+                bit_list = [operation.measure.qubit]
+            else:
+                assert operation.target is not None
+                bit_list = [operation.target]
         elif isinstance(operation, qasm3_ast.QuantumPhase) and operation.qubits is None:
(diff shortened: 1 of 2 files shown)


ryanhill1 (approved):
Nice! After this is merged, we'll wanna update https://github.com/qBraid/pyqasm/pull/122 accordingly, as well.

TheGupta2012 (approved):
Looks good @arulandu !

PR #156, opened 2025-03-03
Implemented Solovay Kitaev's Algorithm

PranavTupe2000:
## Summary of changes
This PR contains implementation of a Solovay Kitaev's Algorithm to estimate conversion of parameterized gates to non parameterized gate to an arbitrary accuracy.


src/pyqasm/algorithms/solovay_kitaev/basic_approximation.py
```diff
@@ -0,0 +1,52 @@
...
+
+    for gate in gate_list:
+        gate_matrix = gate["matrix"]
+        tree_depth = gate["depth"]
+
+        # Stop if the maximum depth is exceeded
+        if tree_depth > max_tree_depth:
+            break
+
+        trace_diff = np.abs(np.trace(np.dot(gate_matrix.conj().T, U) - np.identity(2)))
```
TheGupta2012, line 36: Let us use he operator norm to find the distance between the matrices - 
**Operator Norm** - `np.linalg.norm(gate_matrix - V, 2)`

This is because it is mentioned in the paper that the operator norm is used as the metric to consider the closeness of any approximation

src/pyqasm/algorithms/solovay_kitaev/solovay_kitaev.py
```diff
@@ -0,0 +1,147 @@
...
+    return result, final_error
+
+def solovay_kitaev(target: np.ndarray, target_basis_set, depth: int = 3) -> List[np.ndarray]:
+    """
+    Main function to run the Solovay-Kitaev algorithm.
+    
+    Args:
+        target: Target unitary matrix as numpy array
+        target_basis_set: The target basis set to rebase the module to.
+        depth: Recursion depth
```
TheGupta2012, line 111: Can we try and find some sort of relation between the recursion depth and the depth of the final sequence of gates? It'll be helpful for the user to set a particular depth limit for a gate decomposition

src/pyqasm/algorithms/solovay_kitaev/solovay_kitaev.py
```diff
@@ -0,0 +1,147 @@
...
+    
+    error = target * prev_sequence.dagger()
+    
+    # Find Va and Vb such that their group commutator approximates the error
+    best_v = None
+    best_w = None
+    best_error = float('inf')
+
+    for v in basic_gates:
+        for w in basic_gates:
```
TheGupta2012, line 83: Why do we only consider the `basic_gates` for the error approximation?

src/pyqasm/algorithms/solovay_kitaev/optimizer.py
```diff
@@ -0,0 +1,76 @@
...
+        "tdg": {
+            "group": "sdg-tdg",
+            "weight": 0.125
+        },
+    }
+}
+
+def optimize_gate_sequnce(seq: list[str], target_basis_set):
+    target_identity_weight_group = IDENTITY_WEIGHT_GROUP[target_basis_set]
+    while True:
```
TheGupta2012, line 31: Let's have a finite limit for this processing, we don't wanna get stuck in the loop. `1e6` iterations should be a good limit to consider

src/pyqasm/algorithms/solovay_kitaev/optimizer.py
```diff
@@ -0,0 +1,76 @@
...
+            return seq    
+
+if __name__ == '__main__':
+    s1 = ['s', 's', 's', 't', 't', 'tdg', 'sdg', 'sdg', 'sdg', 'tdg', 's', 'h', 's']
+    s2 = ['t', 's', 's', 's', 't', 'tdg', 'tdg', 'sdg', 'sdg', 'sdg', 't', 's', 's', 's', 't', 'tdg', 'tdg', 'sdg', 'sdg', 'sdg', 's', 'h', 's']
+    s3 = ['h', 's', 's', 't', 't', 's', 't'] # ['h', 't']
+    s4 = ['h', 's', 's', 't', 't', 's', 'h'] # []
+    s5 = ['h', 's', 's', 't', 'h', 'h', 't', 's', 'h', 't'] # ['t']
+    
+    print(optimize_gate_sequnce(s1, BasisSet.CLIFFORD_T) == ['s', 'h', 's'])
```
TheGupta2012, line 72: Can we add some randomized tests for this optimizer? I'd like to see how it performs with random inputs by comparing the matrices of the input and output sequences.

src/pyqasm/algorithms/solovay_kitaev/optimizer.py
```diff
@@ -0,0 +1,76 @@
...
+            "weight": 0.25
+        },
+        "tdg": {
+            "group": "sdg-tdg",
+            "weight": 0.125
+        },
+    }
+}
+
+def optimize_gate_sequnce(seq: list[str], target_basis_set):
```
TheGupta2012, line 29:
```suggestion
def optimize_gate_sequence(seq: list[str], target_basis_set):
```

src/pyqasm/algorithms/solovay_kitaev/optimizer.py
```diff
@@ -0,0 +1,76 @@
...
+    target_identity_weight_group = IDENTITY_WEIGHT_GROUP[target_basis_set]
+    while True:
+        current_group = None
+        current_weight = 0
+        start_index = 0
+        changed = False
+
+        for i, gate_name in enumerate(seq):
+            gate = target_identity_weight_group[gate_name]
```
TheGupta2012, line 38: add a check for the existence of `gate` in the `arget_identity_weight_group` dict

src/pyqasm/algorithms/solovay_kitaev/generator.py
```diff
@@ -0,0 +1,102 @@
+import pickle
+import sys
+from collections import deque
+
+import numpy as np
+
+gate_sets = {
```
TheGupta2012, line 7: Can we instead expand the `IDENTITY_WEIGHT_GROUP` to also contain the matrix and use that here as I see duplication between the two data structures

TheGupta2012 (changes requested):
Really great work on the implementation @PranavTupe2000 ! I've given some comments for the next steps as you start integrating this into the code base.

src/pyqasm/algorithms/solovay_kitaev/optimizer.py
TheGupta2012: Also change the reference at other places

PranavTupe2000:
For now, I have just pushed the code of Traversal basic approximation, if anyone wants to check it out. Now I will integrate it with main code, and then I will perform the testing.

TheGupta2012:
The changes look great @PranavTupe2000! I think the PR is quite close in terms of implementation. I'll add some more comments related to the general code structure. I think the major remaining chunk of implementation is with respect to the tests for performance and accuracy for decomposing non-clifford 1 qubit gates.

src/pyqasm/algorithms/solovay_kitaev/__init__.py
TheGupta2012, line 1: Let us export the algorithm in the `__init__.py` so that we can import it something like - 

```python
from pyqasm.algorithms import solovay_kitaev
``` 

Although we are majorly gonna be using it internally, it helps to import the core functionality of a module for easier imports . See the [`pyqasm/modules`](https://github.com/qBraid/pyqasm/blob/8a04d14949e4ab8adcd2cdb6a1f6dc319571ce9f/src/pyqasm/modules/__init__.py) for reference.

src/pyqasm/decomposer.py
```diff
@@ -159,3 +170,70 @@ def _get_qubits_for_gate(cls, qubits, rule):
...
+                    name=qasm3_ast.Identifier(name=approximated_gate_name),
+                    arguments=[],
+                    qubits=statement.qubits,
+                )
+
+                processed_gates_list.append(new_gate)
+        return processed_gates_list
+
+    @classmethod
+    def _get_target_matrix_for_rotational_gates(cls, gate_name, theta):
```
TheGupta2012, line 214: I believe this method / mapping should be present in the `pyqasm/maps/gates.py` file as it is unrelated to the `Decomposer` logic.

src/pyqasm/maps/decomposition_rules.py
```diff
@@ -96,6 +96,67 @@ class AppliedQubit(Enum):
     },
 }
 
+ROTATIONAL_LOOKUP_RULES = {
+    BasisSet.CLIFFORD_T: {
+        "rz": {
+            CONSTANTS_MAP["pi"]: [
+                {"gate": "s"},
```
TheGupta2012, line 103: Is it necessary to add a "gate" key in this data structure? What about defining the list something like - 
```python
...
"rx": {
            CONSTANTS_MAP["pi"]: [
                "h",
                "s",
                "s",
                "h",
            ]
...
```
 I think it should simplify implementation a little bit.

src/pyqasm/maps/gates.py
```diff
@@ -1111,6 +1111,41 @@ def two_qubit_gate_op(
     BasisSet.CLIFFORD_T: {"h", "t", "s", "cx", "tdg", "sdg"},
 }
 
+GATE_ENTITY_DATA = {
+    BasisSet.CLIFFORD_T: [
+        {
+            "name": "h",
+            "identity": {"group": "h", "weight": 0.5},
```
TheGupta2012, line 1118: can you add a note above the key about what `"identity"` means ?

src/pyqasm/maps/gates.py
```diff
@@ -1111,6 +1111,41 @@ def two_qubit_gate_op(
     BasisSet.CLIFFORD_T: {"h", "t", "s", "cx", "tdg", "sdg"},
 }
 
+GATE_ENTITY_DATA = {
```
TheGupta2012, line 1114: Why have we chosen `ENTITY` here? I think we can also use something like `GATE_OPT_DATA` (short for OPTIMIZATION)

src/pyqasm/decomposer.py
```diff
@@ -159,3 +170,70 @@ def _get_qubits_for_gate(cls, qubits, rule):
...
+            target_basis_set: The target basis set to rebase the module to.
+            depth: The depth of the approximation.
+            accuracy: The accuracy of the approximation.
+            
+        Returns:
+            list: The processed gates based on the target basis set.
+        """
+        theta = statement.arguments[0].value
+        processed_gates_list = []
+
```
TheGupta2012, line 190: Is `_process_rotational_gate` invoked in other basis sets also?

src/pyqasm/algorithms/solovay_kitaev/utils.py
```diff
@@ -0,0 +1,168 @@
...
+        return TU2Matrix(matrix, name, identity_group, identity_weight)
+
+    def can_multiple(self, other: "TU2Matrix"):
+        """Checks if the two matrices can be multiplied.
+        If the identity group is the same, the identity weight should be less than 1.
+        """
+        if self.identity_group != other.identity_group:
+            return True
+
+        if self.identity_weight + other.identity_weight < 1:
```
TheGupta2012, line 105:
```suggestion
        return self.identity_weight + other.identity_weight < 1
```

TheGupta2012, line 105:
```suggestion
        return self.identity_weight + other.identity_weight < 1:
```

TheGupta2012: Can simplify this check

src/pyqasm/algorithms/solovay_kitaev/basic_approximation.py
```diff
@@ -0,0 +1,102 @@
...
+
+    Returns:
+        float: The closest difference between the target and approximated matrix.
+        TU2Matrix: The closest approximated matrix.
+        TU2Matrix: The best approx
+
+    """
+    accuracy, max_tree_depth, closest_diff, closest_gate, best_gate = params
+
+    if current_depth >= max_tree_depth:
```
TheGupta2012, line 30:
```suggestion
    if current_depth >= max_tree_depth or best_gate:
```

src/pyqasm/algorithms/solovay_kitaev/basic_approximation.py
```diff
@@ -0,0 +1,102 @@
...
+        return closest_diff, closest_gate, best_gate
+
+    if best_gate:
```
TheGupta2012, line 33: can combine with above

src/pyqasm/algorithms/solovay_kitaev/basic_approximation.py
```diff
@@ -0,0 +1,102 @@
...
+        closest_diff, closest_gate, best_gate = rescursive_traversal(
+            target_matrix,
+            approximated_matrix.copy(),
+            target_gate_set_list,
+            current_depth + 1,
+            (accuracy, max_tree_depth, closest_diff, closest_gate, best_gate),
+        )
+        approximated_matrix = approximated_matrix_copy.copy()
+
+    return closest_diff, closest_gate, best_gate
```
TheGupta2012, line 61: Should we not return only a single gate i.e. the `best_gate`. I agree that the accuracy might not go below the target accuracy, in which case `closest_gate` becomes the `best_gate`

src/pyqasm/decomposer.py
PranavTupe2000: It will be invoked while rebasing, if QASM code contains a rotational gate, and that gate is not present in target basis set.

src/pyqasm/algorithms/solovay_kitaev/__init__.py
TheGupta2012, line 1: Let's add an import here - 
```python
from .solovay_kitaev import solovay_kitaev
```

and then in the `algorithms` init, use - 

```python
from pyqasm.algorithms.solovay_kitaev import solovay_kitaev
__all__ = ["solovay_kitaev"]
```

Looks a little cleaner that way

src/pyqasm/algorithms/__init__.py
```diff
@@ -0,0 +1,14 @@
+"""
+Sub module for quantum algorithms.
+
+Functions:
+----------
+    solovay_kitaev: Solovay-Kitaev algorithm for approximating unitary gates.
+
+"""
+
+from pyqasm.algorithms.solovay_kitaev.solovay_kitaev import solovay_kitaev
```
ryanhill1, line 10:
```suggestion
from .solovay_kitaev.solovay_kitaev import solovay_kitaev
```

ryanhill1 (changes requested):

src/pyqasm/algorithms/solovay_kitaev/optimizer.py
```diff
@@ -0,0 +1,94 @@
...
+    ]
+    s3 = ["h", "s", "s", "t", "t", "s", "t"]  # ['h', 't']
+    s4 = ["h", "s", "s", "t", "t", "s", "h"]  # []
+    s5 = ["h", "s", "s", "t", "h", "h", "t", "s", "h", "t"]  # ['t']
+
+    print(optimize_gate_sequence(s1, BasisSet.CLIFFORD_T) == ["s", "h", "s"])
+    print(optimize_gate_sequence(s2, BasisSet.CLIFFORD_T) == ["s", "h", "s"])
+    print(optimize_gate_sequence(s3, BasisSet.CLIFFORD_T) == ["h", "t"])
+    print(optimize_gate_sequence(s4, BasisSet.CLIFFORD_T) == [])
+    print(optimize_gate_sequence(s5, BasisSet.CLIFFORD_T) == ["t"])
```
ryanhill1, line 94: Remove

src/pyqasm/algorithms/solovay_kitaev/solovay_kitaev.py
```diff
@@ -0,0 +1,182 @@
...
+    # best_error = float("inf")
+
+    # for v in basic_gates:
+    #     for w in basic_gates:
+    #         comm = group_commutator(v, w)
+    #         curr_error = error.distance(comm)
+    #         if curr_error < best_error:
+    #             best_error = curr_error
+    #             best_v = v
+    #             best_w = w
```
ryanhill1, line 100: Remove if not relevant, or convert to code

src/pyqasm/algorithms/solovay_kitaev/solovay_kitaev.py
```diff
@@ -0,0 +1,182 @@
...
+#     print(np.allclose(r1.matrix, r2.matrix))  # Output: True
+#     print(np.allclose(r2.matrix, r0.matrix))  # Output: True
+
+if __name__ == '__main__':
+    U = np.array([[-0.8987577 +0.j, -0.31607208+0.30386352j],[-0.14188982-0.41485163j,  0.79903828+0.41146473j]])
+    r0 = solovay_kitaev(U, BasisSet.CLIFFORD_T, depth=2)
+    print("----------------")
+    print(r0.distance(SU2Matrix(U, [])))
+    print(r0.matrix)
+    # print(r0.name)
```
ryanhill1, line 182: Remove

src/pyqasm/algorithms/solovay_kitaev/testing.ipynb
ryanhill1, line 1: Remove

src/pyqasm/algorithms/solovay_kitaev/gc_decompose.py
```diff
@@ -0,0 +1,59 @@
...
+    phi = 2.0 * np.arcsin(np.sqrt(np.sqrt((0.5 - 0.5 * np.cos(theta / 2)))))
+    v = RotationX(phi)
+    if axis[2] > 0:
+        w = RotationY(2 * np.pi - phi)
+    else:
+        w = RotationY(phi)
+    
+    _, ud = np.linalg.eig(u)
+    
+    vwvdwd = v @ w @ v.conj().T @ w.conj().T
```
TheGupta2012, line 53: I believe this also requires diagonalization of the matrix as done in a [reference implementation](https://github.com/qcc4cp/qcc/blob/main/src/solovay_kitaev.py#L99).

TheGupta2012: Hi @PranavTupe2000 checking in to see if were you able to solve this issue?

PR #171, opened 2025-04-11
Improve error reporting for pyqasm

TheGupta2012:
## Summary of changes
- Improve the error messages across the repository and reduce traceback size 
- Added a flag `expand_traceback` to `validate` and `unroll` commands. Set to `False` by default, it collapses the traceback to only display the last error msg. If set to `True`, full traceback displayed 
- Added exact lines for QASM errors, eg. - 

```python
import pyqasm
qasm = """
OPENQASM 3.0;
include "stdgates.inc";

def test(qubit[2] q2, int[32] b) -> bool{
    h q2[0];
    cx q2[0], q2[1];
    return true;
}
qubit[10] q;
bool b = test(a, 5);

"""
program = pyqasm.loads(qasm)
program.unroll(expand_traceback=0)
```
Output - 
```bash
ERROR:pyqasm: Error at line 11, column 9 in QASM file

 >>>>>> test(a, 5)

pyqasm.exceptions.ValidationError: Expecting qubit argument for 'q2'. Qubit register 'a' not found for function 'test'
Usage: test ( qubit[2] q2 , int[32] b )
```

### DIFF
--- a/src/pyqasm/exceptions.py
+++ b/src/pyqasm/exceptions.py
@@ -81,10 +81,10 @@ def raise_qasm3_error(
                 error_parts.append("\n >>>>>> " + dumps(error_node, indent="    ") + "\n")
             elif isinstance(error_node, list):
                 error_parts.append(
-                    "\n >>>>>> "
-                    + " , ".join(dumps(node, indent="    ") for node in error_node + "\n")
+                    "\n >>>>>> " + " , ".join(dumps(node, indent="    ") for node in error_node)
                 )
         except Exception as _:  # pylint: disable = broad-exception-caught
+            print(_)
             error_parts.append("\n >>>>>> " + str(error_node))
 
     if error_parts:
--- a/src/pyqasm/exceptions.py
+++ b/src/pyqasm/exceptions.py
@@ -84,7 +84,13 @@ def raise_qasm3_error(
 
     if error_node:
         try:
-            error_parts.append("\n >>>>>> " + dumps(error_node, indent="    "))
+            if isinstance(error_node, QASMNode):
+                error_parts.append("\n >>>>>> " + dumps(error_node, indent="    ") + "\n")
+            elif isinstance(error_node, list):
+                error_parts.append(
+                    "\n >>>>>> "
+                    + " , ".join(dumps(node, indent="    ") for node in error_node + "\n")
+                )
         except Exception as _:  # pylint: disable = broad-exception-caught
             error_parts.append("\n >>>>>> " + str(error_node))
 
--- a/src/pyqasm/exceptions.py
+++ b/src/pyqasm/exceptions.py
@@ -17,7 +17,6 @@ Module defining base PyQASM exceptions.
 
 """
 
-import logging
 import os
 import sys
 from typing import Optional, Type
@@ -26,13 +25,7 @@ from openqasm3.ast import QASMNode, Span
 from openqasm3.parser import QASM3ParsingError
 from openqasm3.printer import dumps
 
-# Define a custom logger for the module
-handler = logging.StreamHandler()
-handler.setFormatter(logging.Formatter("%(levelname)s:%(name)s: %(message)s"))
-
-logger = logging.getLogger("pyqasm")
-logger.addHandler(handler)
-logger.setLevel(logging.WARNING)
+from ._logging import logger
 
 
 class PyQasmError(Exception):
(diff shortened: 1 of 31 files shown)


src/pyqasm/maps/expressions.py
ryanhill1, line 1: Why putting back Unions?

src/pyqasm/exceptions.py
```diff
@@ -56,17 +68,39 @@ def raise_qasm3_error(
...
+                    "\n >>>>>> "
+                    + " , ".join(dumps(node, indent="    ") for node in error_node + "\n")
+                )
+        except Exception as _:  # pylint: disable = broad-exception-caught
+            error_parts.append("\n >>>>>> " + str(error_node))
+
+    if error_parts:
+        logger.error("\n".join(error_parts))
+
+    if os.environ.get("PYQASM_EXPAND_TRACEBACK") == "0":
```
ryanhill1, line 100: if named `PYQASM_EXPAND_TRACEBACK`, feels more like it should be set to a "true" or "false". Setting it equal to a number e.g. "0" or "1" would make more sense if this environment variable was named something like`PYQASM_TRACEBACKLIMIT` for consistency.

ryanhill1 (changes requested):

src/pyqasm/maps/expressions.py
TheGupta2012: My bad, requires `rebase` from `main`

src/pyqasm/exceptions.py
```diff
@@ -18,10 +18,21 @@
...
 from openqasm3.parser import QASM3ParsingError
+from openqasm3.printer import dumps
+
+# Define a custom logger for the module
+handler = logging.StreamHandler()
+handler.setFormatter(logging.Formatter("%(levelname)s:%(name)s: %(message)s"))
+
+logger = logging.getLogger("pyqasm")
+logger.addHandler(handler)
+logger.setLevel(logging.WARNING)
```
ryanhill1, line 35: We want to be careful that doing this doesn't cause the user to not be able to modify the logging cofigs themselves. For example, if a user wanted to later on set the log level to something different, they should still be able to do that. See https://github.com/qBraid/qBraid/issues/941 as an example where we had a similar logging config hard-coded in the SDK which caused an inconvenience to an end-user.

src/pyqasm/modules/base.py
```diff
@@ -504,11 +505,12 @@ def reverse_qubit_order(self, in_place=True):
         # 4. return the module
         return qasm_module
 
-    def validate(self):
+    def validate(self, expand_traceback: Optional[bool] = False):
         """Validate the module"""
         if self._validated_program is True:
             return
         try:
+            os.environ["PYQASM_EXPAND_TRACEBACK"] = "true" if expand_traceback else "false"
```
ryanhill1, line 513: Dynamically setting environment variables like this isn't good practice, especially when it's done implicitly. Why can't we just do something like:

```python
env_expand = os.getenv("PYQASM_EXPAND_TRACEBACK", "false").lower() == "true"
expand_traceback = expand_traceback if expand_traceback is not None else env_expand
```

and just pass around a python variable rather getting and setting values from environment variables?

src/pyqasm/modules/base.py
```diff
@@ -538,6 +541,10 @@ def unroll(self, **kwargs):
         if not kwargs:
             kwargs = {}
         try:
+            os.environ["PYQASM_EXPAND_TRACEBACK"] = (
+                "true" if kwargs.pop("expand_traceback", False) else "false"
+            )
```
ryanhill1, line 546: same comment as above

ryanhill1 (changes requested):

src/pyqasm/modules/base.py
TheGupta2012: The main motivation for setting this env variable was to allow [`raise_qasm3_error`](https://github.com/qBraid/pyqasm/pull/171/files#diff-441d11894c4665cd20dc9be68f23c5780151c13bea7a07faf3332ec5411be56eR100) function to read the value and determine verbosity of the traceback. This eliminated the need for passing the `expand_traceback` as a parameter.

I could easily have made it as an attribute of the `QasmVisitor` class but then we would need to pass it around in every call to the exception raising method. 

Also, I think we can set the variable during initialization of the `pyqasm` package. In that way, users can explicitly choose to change the variable on their own and we'll just use the `getenv` method ( as suggested above ).

ryanhill1 (approved):

ryanhill1 (dismissed):

PR #180, opened 2025-04-30
Remove deepcopy in `_visit_generic_gate_operation`

TheGupta2012:
Reference : [#962 | qBraid](https://github.com/qBraid/qBraid/issues/962)

## Summary of changes
The change changes the deep copy of the `operation` and `ctrls` variables into a shallow copy, likely to improve performance and reduce unnecessary memory usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Types of changes:
 
 ### Improved / Modified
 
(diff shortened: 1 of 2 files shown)


TheGupta2012:
@bachase post this change can you test the benchmarks against the `main` branch of the repo? 
I believe that the shallow `copy.copy` operation would be less expensive than the `deepcopy` but happy to see the test results and comment further.

bachase:
@TheGupta2012 a nice improvement! 

The call stack starting from https://github.com/qBraid/qBraid/blob/7ac709340740e6703acb81cbee586112573d703b/qbraid/transpiler/conversions/qasm2/qasm2_to_cirq.py#L50 went from 13.67 seconds to 3.16 seconds.

PR #186, opened 2025-05-23
Launch on qBraid badge color

ryanhill1:
## Summary of changes

- Change badge color to light grey instead of white so that it is more clearly outlined


ryanhill1:
Nope that was it

Issue #191, opened 2025-05-26
Refactor `Qasm3Analyzer.analyze_classical_indices` to `@staticmethod`


TheGupta2012:
### Description
The [Qasm3Analyzer.analyze_classical_indices method](https://github.com/qBraid/pyqasm/blob/a3ba3e80e9b1de6143441c7c68eca783a62dd1e2/src/pyqasm/analyzer.py#L50) is currently defined as a `@classmethod`, but it does not use the `cls` parameter or any class attributes. It should be refactored to a `@staticmethod` for clarity and consistency with the rest of the utility methods.

### Tasks

- Change the decorator from @classmethod to @staticmethod.
- Remove the cls parameter from the method signature.
- Ensure the method and its usage do not depend on class-level attributes.
- Update any code/tests that call this method if required.

samuraikillers:
Hi @TheGupta2012 I will work on the issue

TheGupta2012:
Sure, have a go at it!

samuraikillers:
Hi @TheGupta2012 please review the PR #194 , closed the first one due to some mistakes :)

PR #192, opened 2025-05-26
changing class to static method

samuraikillers:
Fix for issue #191

### DIFF
--- a/src/pyqasm/expressions.py
+++ b/src/pyqasm/expressions.py
@@ -155,7 +155,7 @@ class Qasm3ExprEvaluator:
             var_value = cls.visitor_obj._get_from_visible_scope(var_name).value
         else:
             validated_indices = Qasm3Analyzer.analyze_classical_indices(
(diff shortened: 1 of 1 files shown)


src/pyqasm/expressions.py
```diff
@@ -155,7 +155,7 @@ def _get_var_value(cls, var_name, indices, expression):
             var_value = cls.visitor_obj._get_from_visible_scope(var_name).value
         else:
             validated_indices = Qasm3Analyzer.analyze_classical_indices(
-                indices, cls.visitor_obj._get_from_visible_scope(var_name), cls
+                indices, cls.visitor_obj._get_from_visible_scope(var_name), Qasm3ExprEvaluator
```
TheGupta2012, line 158: You don't need to update the `cls` here as we are referring to the current expression evaluator object.

samuraikillers: @TheGupta2012 sure will revert back

PR #198, opened 2025-05-29
Fix depth calculation for external gates

### LINKED ISSUE #77: [BUG] Depth calculation for external gates
### Environment

- **PyQASM version**: source
- **Python version**: 3.12.7
- **Operating system**: macOS 14.4

### What happened?

With the merging of #59 , we need to update the `depth` calculation of the `QasmModule`. As mentioned in the PR, when a module is unrolled with external gates, the depth calculation implicitly refers the gate definition to calculate the final depth. See [unrolling example in external gates PR](https://github.com/qBraid/pyqasm/pull/59#issuecomment-2463864045).

### Suggestions (Optional)

- As stated, we can treat any `external_gate` as "atomic" whenever we are calculating the depth of the quantum program. Moreover, we need to update the bit depths inside the `_visit_external_gate_operation` method after [broadcasting it](https://github.com/qBraid/pyqasm/blob/main/pyqasm/visitor.py#L869) on the qubits arguments. 

- Note that the validation steps for custom gate and basic gate already have depth calculations inside their definitions. We'll either need to _disable them_ OR think of _how we want to define depth_ in the first place. 
- By the latter, I mean that do we want to include gate definitions of external gates in depth calculations or not? Since we  force the gate to be defined (by the user or as a basic gate), we are specifying its depth through the definition. 

There are two approaches to handle this - 
1. **Depth is dependent on external gate definitions**: We calculate the `depth` of the circuit as defined by the user gate definitions. Even though a person specifies an external gate, the depth is calculated by considering its definition - 

```python
In [6]: qasm_str = """
   ...: OPENQASM 3.0;
   ...: include "stdgates.inc";
   ...: gate custom_native_gate q1, q2 {
   ...:   h q1;
   ...:   cx q1, q2;
   ...:   h q2;
   ...: }
   ...: qubit[2] q;
   ...: custom_native_gate q[0], q[1];
   ...: """

In [7]: mod = pyqasm.loads(qasm_str)

In [8]: mod.unroll(external_gates = ["custom_native_gate"])

In [9]: mod.depth()
Out[9]: 3
``` 
A corner case is when the body is empty. Here the depth should be incremented by 1 but because of the empty body, it will just be 0 - 

```python
In [1]: import pyqasm

In [2]: qasm_str = """
   ...: OPENQASM 3.0;
   ...: include "stdgates.inc";
   ...: gate custom_native_gate q1, q2 { }
   ...: qubit[2] q;
   ...: custom_native_gate q[0], q[1];
   ...: """

In [3]: mod = pyqasm.loads(qasm_str)

In [4]: mod.unroll(external_gates = ["custom_native_gate"])

In [5]: mod.depth()
Out[5]: 0
``` 
In this case, we will add 1 to the depth while calculating it inside the external gate handler. 

2. **Depth is independent of external definitions**: We ignore the external gates' definitions and treat them as gates with empty body. This means - 

```c
gate custom_native_gate q1, q2 {
   h q1;
   cx q1, q2;
   h q2;
 }
```

is treated as - 

```c
gate custom_native_gate q1, q2 { }
``` 
This will not affect the depth calculations as there won't be any statement to visit while processing this external gate. Then, at the end, we just update the depths of qubits appropriately.

Happy to discuss this more @Sola85 and @ryanhill1

antalszava:
## Summary of changes

This PR fixes how the depth of a QASM module is calculated when calling the `depth` method.

With the changes in this PR, a QASM module stores external gates that have been defined for it. At the moment, the main way to define external gates on a module is to pass the `external_gates` keyword argument to the `unroll` method. Calling the same `unroll` method with no such arguments being passed flushes the external gates of the module.

Storing such external gates on the module level helps the depth calculation: any external gates of the module contribute once to the overall depth of the module, instead of the definition of the external gate being considered.

Closes #77.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ Types of changes:
 
 ### Fixed
 
+- Fixed the way how depth is calculated when external gates are defined with unrolling a QASM module. ([#XXX](https://github.com/qBraid/pyqasm/pull/XXX))
+
 ### Dependencies
 
 ### Other
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ Types of changes:
 
 ### Fixed
 
-- Fixed the way how depth is calculated when external gates are defined with unrolling a QASM module. ([#XXX](https://github.com/qBraid/pyqasm/pull/XXX))
+- Fixed the way how depth is calculated when external gates are defined with unrolling a QASM module. ([#198](https://github.com/qBraid/pyqasm/pull/198))
 
 ### Dependencies
 
--- a/src/pyqasm/modules/base.py
+++ b/src/pyqasm/modules/base.py
@@ -55,6 +55,7 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
         self._has_barriers: Optional[bool] = None
         self._validated_program = False
         self._unrolled_ast = Program(statements=[])
+        self._external_gates = []
 
     @property
     def name(self) -> str:
@@ -278,7 +279,10 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
         qasm_module = self.copy()
         qasm_module._qubit_depths = {}
         qasm_module._clbit_depths = {}
-        qasm_module.unroll()
+
+        # Unroll using any external gates that have been recorded for this
+        # module
+        qasm_module.unroll(external_gates = self._external_gates)
 
         max_depth = 0
         max_qubit_depth, max_clbit_depth = 0, 0
@@ -539,6 +543,10 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
             kwargs = {}
         try:
             self.num_qubits, self.num_clbits = 0, 0
+            if ext_gates := kwargs.get("external_gates"):
+                self._external_gates = ext_gates
+            else:
+                self._external_gates = []
             visitor = QasmVisitor(module=self, **kwargs)
             self.accept(visitor)
         except (ValidationError, UnrollError) as err:
--- a/src/pyqasm/modules/base.py
+++ b/src/pyqasm/modules/base.py
@@ -55,7 +55,7 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
         self._has_barriers: Optional[bool] = None
         self._validated_program = False
         self._unrolled_ast = Program(statements=[])
-        self._external_gates = []
+        self._external_gates: list[str] = []
 
     @property
     def name(self) -> str:
--- a/src/pyqasm/modules/base.py
+++ b/src/pyqasm/modules/base.py
@@ -282,7 +282,7 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
 
         # Unroll using any external gates that have been recorded for this
         # module
-        qasm_module.unroll(external_gates = self._external_gates)
+        qasm_module.unroll(external_gates=self._external_gates)
 
         max_depth = 0
         max_qubit_depth, max_clbit_depth = 0, 0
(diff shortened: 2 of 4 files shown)


src/pyqasm/modules/base.py
```diff
@@ -278,7 +279,10 @@ def depth(self):
         qasm_module = self.copy()
         qasm_module._qubit_depths = {}
         qasm_module._clbit_depths = {}
-        qasm_module.unroll()
+
+        # Unroll using any external gates that have been recorded for this
+        # module
+        qasm_module.unroll(external_gates = self._external_gates)
```
antalszava, line 285: Note: the design of this solution has been greatly influenced by this internal call: even though `unroll` may be called externally with the `external_gates` kwarg defined, internally here it has always been called without. Imho this can be problematic, hence the current approach that is being proposed.

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ Types of changes:
 
 ### Fixed
 
+- Fixed the way how depth is calculated when external gates are defined with unrolling a QASM module. ([#XXX](https://github.com/qBraid/pyqasm/pull/XXX))
```
TheGupta2012, line 31:
```suggestion
- Fixed the way how depth is calculated when external gates are defined with unrolling a QASM module. ([#198](https://github.com/qBraid/pyqasm/pull/198))
```

src/pyqasm/visitor.py
```diff
@@ -760,12 +761,13 @@ def _update_qubit_depth_for_gate(
                 qubit_node.num_gates += 1
```
TheGupta2012, line 761: Say we have the following program- 
```c
OPENQASM 3.0;
include "stdgates.inc";
gate my_gate() q {
 h q;
 x q;
}
qubit[2] q;
my_gate q[0];
``` 
According to the above logic, we shall have `num_gates` for qubit 0 as `3` whereas its depth would be `1` - 

```python
In [17]: import pyqasm

In [18]: qasm_str = """
    ...: OPENQASM 3.0;
    ...: include "stdgates.inc";
    ...: gate my_gate() q {
    ...:  h q;
    ...:  x q;
    ...: }
    ...: qubit[2] q;
    ...: my_gate q[0];"""

In [19]: mod = pyqasm.loads(qasm_str)

In [20]: mod.unroll(external_gates = ['my_gate'])

In [21]: mod._qubit_depths
Out[21]: 
{('q',
  0): QubitDepthNode(reg_name='q', reg_index=0, depth=1, num_resets=0, num_measurements=0, num_gates=3, num_barriers=0),
 ('q',
  1): QubitDepthNode(reg_name='q', reg_index=1, depth=0, num_resets=0, num_measurements=0, num_gates=0, num_barriers=0)}
```

This is happening because we update the `num_gates` both inside the two basic gates `h` and `x`, and while exiting the handling of the external gate itself! This property is [used in the printer](https://github.com/qBraid/pyqasm/blob/5d48d7f78d89507e9c95a27f203c434b087b14df/src/pyqasm/printer.py#L180) but the regression isn't detected as we don't have a visualization test with external gates. 

Ideally, the check for `self._recording_depth` should be moved over the whole gate and depth update logic

src/pyqasm/visitor.py
```diff
@@ -88,6 +88,7 @@ def __init__(
         self._unroll_barriers: bool = unroll_barriers
         self._curr_scope: int = 0
         self._label_scope_level: dict[int, set] = {self._curr_scope: set()}
+        self._recording_depth = True
```
TheGupta2012, line 91: Let's name this to `self._recording_ext_gate_depth` as `recording_depth` sounds very general.
Then, `True` will mean we are inside the external gate and `False` otherwise. Essentially, we just flip your current logic

src/pyqasm/modules/base.py
TheGupta2012: I guess `external_gates` is the only kwarg affecting the depth calculation so it is okay to use that here.

antalszava:
Hi @TheGupta2012, thank you! I have made updates as per the suggestions.

TheGupta2012 (approved):
Changes lgtm, thanks for working on this @antalszava !

PR #199, opened 2025-05-30
feat: Add support for while statement

### LINKED ISSUE #7: [FEATURE] Add support for while statements
### Feature Description

Add support for the [`while` loops in openqasm](https://openqasm.com/language/classical.html#while-loops). The main motivation behind this feature is to provide complete semantic coverage of the [openqasm3 grammar](https://openqasm.com/language/index.html)

### Current 
The following code containing a `while` loop in openqasm fails during `unroll` - 

```python
import pyqasm

qasm_str = """
OPENQASM 3.0;
include "stdgates.inc";
qubit q;
bit result;

int i = 0;
// Keep applying hadamards and measuring a qubit
// until 10, |1>s are measured
while (i < 10) {
    h q;
    result = measure q;
    if (result) {
        i += 1;
    }
}"""

module = pyqasm.loads(qasm_str)

module.unroll()
```

```output 
...
ValidationError: Unsupported statement of type <class 'openqasm3.ast.WhileLoop'>
```

### Implementation (Optional)

- Similar to the handling of [`for` loop statements](https://github.com/qBraid/pyqasm/blob/f906f8e0fe22217b03b841a717aa804f9e492279/src/pyqasm/visitor.py#L1595), we will need to add implementation for the [`_visit_while_loop`](https://github.com/qBraid/pyqasm/blob/f906f8e0fe22217b03b841a717aa804f9e492279/src/pyqasm/visitor.py#L1781) method in the [`QasmVisitor` class](https://github.com/qBraid/pyqasm/blob/f906f8e0fe22217b03b841a717aa804f9e492279/src/pyqasm/visitor.py#L50)

- This will require analyzing the structure of [`openqasm3.ast.WhileLoop`](https://github.com/openqasm/openqasm/blob/main/source/openqasm/openqasm3/ast.py#L988) class and figuring out how to evaluate the conditions in different iterations.

luffy-orf:
### Summary

This PR adds full support for the `while` statement in OpenQASM 3.0, including correct handling of `break` and `continue` statements within loops. This brings the project closer to complete semantic coverage of the OpenQASM 3 grammar.

### Details

- **Implements `_visit_while_loop` in `QasmVisitor`**:
  - Evaluates the loop condition and executes the loop body as long as the condition is true.
  - Supports a configurable loop iteration limit (default: 1000) to prevent infinite loops, raising a `ValidationError` if exceeded.
  - Handles `break` and `continue` using custom exceptions, ensuring correct control flow.
- **Adds visitor methods for `break` and `continue` statements**:
  - These are now recognized and handled in the AST visitor.
- **Updates the statement visitor dispatch map** to include `WhileLoop`, `BreakStatement`, and `ContinueStatement` nodes.
- **Test coverage**:
  - Adds a test for a `while` loop with both `break` and `continue` in `tests/qasm3/test_loop.py`.
  - Ensures correct unrolling and AST output for OpenQASM 3 code using `while` loops.
  
### Example

```qasm
OPENQASM 3.0;
qubit q;
bit result;
int i = 0;
while (i < 10) {
    h q;
    result = measure q;
    if (result) {
        i += 1;
    }
}
```
- The loop limit is configurable via `self._loop_limit` (default: 1000). Set to `None` to disable.
- The implementation is robust against infinite loops, as discussed in the issue thread.

**Closes #7.**


tests/qasm3/test_loop.py
```diff
@@ -332,3 +332,31 @@ def test_convert_qasm3_for_loop_unsupported_type(caplog):
 
     assert "Error at line 9, column 16" in caplog.text
     assert 'for bit b in "001"' in caplog.text
+
+
+def test_while_loop_with_break_and_continue():
```
TheGupta2012, line 337: Can you add more tests for the change eg. - 
1. Loop limit exceeded exceptions 
2. Comparison of exact qasm strings for unrolled while loops
3. Empty while loops are ignored

src/pyqasm/visitor.py
```diff
@@ -2186,6 +2240,10 @@ def visit_statement(self, statement: qasm3_ast.Statement) -> list[qasm3_ast.Stat
             qasm3_ast.SubroutineDefinition: self._visit_subroutine_definition,
             qasm3_ast.ExpressionStatement: lambda x: self._visit_function_call(x.expression),
             qasm3_ast.IODeclaration: lambda x: [],
+            qasm3_ast.WhileLoop: self._visit_while_loop,
+            # Add break/continue if present in AST
+            getattr(qasm3_ast, 'BreakStatement', None): self._visit_break_statement,
```
TheGupta2012, line 2245: Since `break` and `continue` raise syntax error when existing out of loops, we will never encounter a case where the AST will contain them independently. So, we can remove these checks

src/pyqasm/visitor.py
```diff
@@ -1939,9 +1939,63 @@ def _visit_function_call(
 
         return return_value, result
 
-    def _visit_while_loop(self, statement: qasm3_ast.WhileLoop) -> None:
+    # Custom exceptions for break/continue
+    class BreakException(Exception):
```
TheGupta2012, line 1943: Can we move these into the [`src/pyqasm/exceptions.py`](https://github.com/qBraid/pyqasm/blob/7076d46e057085fbb6189dbe52d6e6ab4b2b2170/src/pyqasm/exceptions.py) ? 
NOTE : these are not errors, so let us not base them with the `PyqasmError` class. Something like `LoopException` would be apt as the base class

src/pyqasm/visitor.py
```diff
@@ -1939,9 +1939,63 @@ def _visit_function_call(
...
+            result.extend(cond_stmts)
+            if not cond_value:
+                break
+            self._push_context(Context.BLOCK)
+            self._push_scope({})
+            self._curr_scope += 1
+            self._label_scope_level[self._curr_scope] = set()
+            try:
+                result.extend(self.visit_basic_block(statement.block))
+            except self.BreakException:
```
TheGupta2012, line 1973: this changes according to the above comment

TheGupta2012 (changes requested):
Thanks @luffy-orf for your work! The changes look good overall, I've added some small comments about refactoring and testing

Have you also thought about how this will handle conditions with qubit measurement results? This is because we can not , at compile time, unroll loops which depend on quantum measurements.

src/pyqasm/visitor.py
```diff
@@ -1939,9 +1939,63 @@ def _visit_function_call(
...
+        raise self.BreakException()
+
+    def _visit_continue_statement(self, statement):
+        """Visit a continue statement element."""
+        raise self.ContinueException()
+
+    def _visit_while_loop(self, statement: qasm3_ast.WhileLoop) -> list[qasm3_ast.Statement]:
+        """Visit a while loop statement element."""
+        result = []
+        loop_limit = getattr(self, "_loop_limit", 1000)
```
TheGupta2012, line 1960:
```suggestion
        loop_limit = getattr(self, "_loop_limit", DEAFULT_MAX_LOOP_ITERATIONS)
```

Also, we'll need to add this attribute to the [visitor class during initialization](https://github.com/qBraid/pyqasm/blob/7076d46e057085fbb6189dbe52d6e6ab4b2b2170/src/pyqasm/visitor.py#L65) and passed in via the module

Sola85:
> Moreover, have you thought about how this will handle conditions with qubit measurement results? This is because we can not unroll loops at compile time which depend on quantum measurements.

I am wondering about this also. In particular, I'm not sure it's feasable to have something like a `LoopLimitExceeded` exception in the general case where loop conditions have to be evaluated at runtime.

TheGupta2012:
> > Moreover, have you thought about how this will handle conditions with qubit measurement results? This is because we can not unroll loops at compile time which depend on quantum measurements.
> 
> I am wondering about this also. In particular, I'm not sure it's feasable to have something like a `LoopLimitExceeded` exception in the general case where loop conditions have to be evaluated at runtime.

We can think of the loop limit variable analogous to the max recursion depth in Python. So in case of a possible infinite loop case this will restrict the unrolled program from taking too much memory. 

Well, a wild suggestion would be to break down the while in terms of nested if statements, whenever we have quantum measurements involved! So something like - 

```c 
while ( c < 1024 ) { // 2^10 
// statements 
// measurements updating the value of bit-register c 
} 
```

will be broken down into - 

```c 
if ( c < 1024 ) { 
    // loop statements 
    if ( c < 1024 ) { 
       // loop statements    
    }
}
```

We can call this something like "fixed iteration unrolling". Although we might not be able to guess when we will get the correct results, but if a QC supports if constructs, this might be one hack to implement while loops conditioned on quantum measurements. 

Also, I might be completely wrong :)

src/pyqasm/visitor.py
```diff
@@ -1939,8 +1943,62 @@ def _visit_function_call(
...
+        result = []
+        loop_limit = getattr(self, "_loop_limit", DEFAULT_MAX_LOOP_ITERATIONS)
+        iterations = 0
+        # Check for quantum measurement in condition (pseudo-code, adjust as needed)
+        if self._condition_depends_on_measurement(statement.test):
+            raise NotImplementedError(
+                "Cannot unroll while loops with quantum measurement in the condition."
+            )
+        while True:
+            cond_value, cond_stmts = Qasm3ExprEvaluator.evaluate_expression(statement.test)
```
atharva-satpute, line 1965: where is the `.test` variable declared?

tests/qasm3/test_loop.py
```diff
@@ -332,3 +332,134 @@ def test_convert_qasm3_for_loop_unsupported_type(caplog):
...
+        for int j in [0, 1] {
+            h q[j];
+        }
+        i += 1;
+    }
+    """
+    result = loads(qasm_str)
+    result.unroll()
+    stmts = [str(s) for s in result.unrolled_ast if hasattr(s, '__str__')]
+    assert sum('h q' in s for s in stmts) == 4
```
TheGupta2012, line 465: please validate qubit indices as well with `check_single_qubit_op` (as you have done in [above test](https://github.com/qBraid/pyqasm/pull/199/files#diff-9188f01a8792f49f21cfe6c2c54a48cb360bd1dc3baff3cf33da5f922789ecccR362)). 

Similar comment applies to all the new tests

src/pyqasm/visitor.py
```diff
@@ -2186,6 +2244,10 @@ def visit_statement(self, statement: qasm3_ast.Statement) -> list[qasm3_ast.Stat
             qasm3_ast.SubroutineDefinition: self._visit_subroutine_definition,
             qasm3_ast.ExpressionStatement: lambda x: self._visit_function_call(x.expression),
             qasm3_ast.IODeclaration: lambda x: [],
+            qasm3_ast.WhileLoop: self._visit_while_loop,
+            # Add break/continue if present in AST
+            getattr(qasm3_ast, "BreakStatement", None): self._visit_break_statement,
+            getattr(qasm3_ast, "ContinueStatement", None): self._visit_continue_statement,
```
TheGupta2012, line 2250: I think we can remove these lines, as I mentioned in an [earlier comment](https://github.com/qBraid/pyqasm/pull/199#discussion_r2115518683)

src/pyqasm/visitor.py
luffy-orf: The .test attribute comes from the OpenQASM 3 AST: WhileLoop.test holds the loop condition, as defined in openqasm3.ast. That’s why we use statement.test here.

src/pyqasm/visitor.py
```diff
@@ -1939,8 +1949,34 @@ def _visit_function_call(
...
+        """Visit a while loop statement element."""
+        result: list[qasm3_ast.Statement] = []
+        # Check for classical register in while condition
+        if self.classical_register_in_expr(statement.while_condition):
+            raise_qasm3_error(
+                "Cannot unroll while loops with quantum measurement in the condition."
+            )
+        iterations = 0
+        loop_limit = self._loop_limit
+        self._curr_scope += 1
```
TheGupta2012, line 1962: Hi @luffy-orf , any reason why you removed [the addition of scopes](https://github.com/qBraid/pyqasm/pull/199/commits/a9b650a958de124bbe12b356d645286936198f38#diff-802ca57af29fc92ca8eb2f536fb5217c61c874c4c35bd39b7fd82512ee0e9a84L1969) from here and the break / continue statements?

src/pyqasm/maps/constants.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Constants for pyqasm maps."""
+
+DEFAULT_MAX_LOOP_ITERATIONS = 1000
```
TheGupta2012, line 17: Let's change this to `1e10` (10 billion), as `1000` iterations are very limiting if not specified by the user

TheGupta2012:
Hi @luffy-orf , any particular reasons to close the above PR?

TheGupta2012:
> > Hi @luffy-orf , any particular reasons to close the above PR?
> 
> Tried fixing the tests but couldnt

Hey, no worries at all! Could you tell me where you're facing issues? I can help out as well. The PR is quite close and once tests are done, it should be good to go

PR #200, opened 2025-05-31
Feature/Added depth calculation for branching statements

### LINKED ISSUE #53: [FEATURE] Update the depth calculation to include branching operations with classical registers
### Feature Description

**pyqasm**: 0.3.2
**qiskit**: 1.2.0 
**qiskit-qasm3-import**: 0.5.1

Quantum programs with branching conditions containing classical registers should be considered for depth calculations. For example, the following circuit - 

```python
In [1]: from qiskit.qasm3 import loads

In [2]: qasm_str = """
   ...: OPENQASM 3.0;
   ...: include "stdgates.inc";
   ...: qreg q[3];
   ...: creg c[3];
   ...: measure q[0] -> c[0];
   ...: cx q[0], q[1];
   ...: measure q[0] -> c[1];
   ...: measure q[0] -> c[2];
   ...: if(c[2]) { h q[2]; }
   ...: """

In [3]: qc = loads(qasm_str)

In [4]: qc.draw(cregbundle = False)
Out[4]: 
     ┌─┐     ┌─┐┌─┐                       
q_0: ┤M├──■──┤M├┤M├───────────────────────
     └╥┘┌─┴─┐└╥┘└╥┘                       
q_1: ─╫─┤ X ├─╫──╫────────────────────────
      ║ └───┘ ║  ║ ┌────── ┌───┐ ───────┐ 
q_2: ─╫───────╫──╫─┤ If-0  ┤ H ├  End-0 ├─
      ║       ║  ║ └──╥─── └───┘ ───────┘ 
c_0: ═╩═══════╬══╬════╬═══════════════════
              ║  ║    ║                   
c_1: ═════════╩══╬════╬═══════════════════
                 ║    ║                   
c_2: ════════════╩════■═══════════════════
                                          

In [5]: qc.depth()
Out[5]: 5
```
has a depth of 5 whereas `pyqasm` gives a depth of 4 - 

```python
In [7]: import pyqasm

In [8]: qasm_str = """
   ...: OPENQASM 3.0;
   ...: include "stdgates.inc";
   ...: qreg q[3];
   ...: creg c[3];
   ...: measure q[0] -> c[0];
   ...: cx q[0], q[1];
   ...: measure q[0] -> c[1];
   ...: measure q[0] -> c[2];
   ...: if(c[2]) { h q[2]; }
   ...: """

In [9]: pyqasm.loads(qasm_str).depth()
Out[9]: 4
```

This is because the classical register depth is not considered during the calculation of qubit depths inside the if block. The depth of `c_2` is 4 when it is used to condition the `H` gate on `q_2`.

### Implementation (Optional)

- While unrolling the branching conditions, we look at whether the condition contains a set of classical register bits or not. We need to update the depth of all these classical bits to the max depth of this set of bits, eg. - 

```python
In [75]: qc.draw(cregbundle=False)
Out[75]:
     ┌─┐     ┌─┐┌─┐
q_0: ┤M├──■──┤M├┤M├─────
     └╥┘┌─┴─┐└╥┘└╥┘
q_1: ─╫─┤ X ├─╫──╫──────
      ║ └───┘ ║  ║ ┌───┐
q_2: ─╫───────╫──╫─┤ H ├
      ║       ║  ║ └─╥─┘
c_0: ═╩═══════╬══╬═══╬══
              ║  ║   ║
c_1: ═════════╩══╬═══╬══
                 ║   ║
c_2: ════════════╩═══■══

c_3: ═══════════════════


In [76]: qc.depth()
Out[76]: 5
```

Here, if `c_0`, `c_1` and `c_2` were used in a branching statement, the depth of each bit would be set to 4 as max depth of bits i.e. `c_2` is 4. In above example, since `c_2` is used for the classically conditioned `H` gate, the depth of the circuit is supposed to be 5 but is coming out as 4 with `pyqasm` - 

```python
In [82]: import pyqasm

In [83]: qasm_str = 'OPENQASM 3.0;\ninclude "stdgates.inc";\nbit[4] c;\nqubit[3] q;\nc[0] = m
    ...: easure q[0];\ncx q[0], q[1];\nc[1] = measure q[0];\nc[2] = measure q[0];\nif (c[2])
    ...: {\n  h q[2];\n}\n'

In [84]: pyqasm.load(qasm_str).depth()
Out[84]: 4
```

- It is not entirely clear how this "max depth of involved classical bits" will be trickled down to the qubits of the if block. A brute force approach is used inside the [qbraid sdk's depth function](https://github.com/qBraid/qBraid/blob/29ec87738d3829169f5421cce0796cdb95bda681/qbraid/passes/qasm/analyze.py#L67) but might not work for a recursive approach used in `QasmVisitor`.

vinayswamik:
## Summary of changes
This PR addresses an issue in `QasmVisitor` where depth calculations for `Gate` and `Measurement` operations inside branching statements were being treated the same as `non-conditional` operations. As a result, both quantum and classical registers were not reflecting the correct depth, leading to inaccurate `circuit-depth`. <br>

With this change, we introduce logic to:

- Detect when we are inside a branching statement `(e.g., if or else)`: we set a flag at the start of `QasmVisitor`.
- Suspend immediate depth updates for all basic gates, custom gates, and measurements while inside the branch. Instead, we collect (and deduplicate) a set of all affected depth nodes (both quantum‐ and classical‐bit nodes) during the entire recursive traversal of that branch.
- Treat the entire branching block as a single depth increase.
- After exiting the branching block, take the maximum depth among all collected nodes and raise each one to “one step above” that maximum. In effect, this synchronizes every `qubit` and `clbit` participating in the conditional arguments to the correct final depth.

Closes #53

### DIFF
--- a/src/pyqasm/visitor.py
+++ b/src/pyqasm/visitor.py
@@ -775,7 +775,8 @@ class QasmVisitor:
                     qubit_node = self._module._qubit_depths[(qubit.name.name, qubit_id)]
                     qubit_node.depth = max_involved_depth
 
-    def _visit_basic_gate_operation(  # pylint: disable=too-many-locals
+    # pylint: disable=too-many-branches, too-many-locals
+    def _visit_basic_gate_operation(
         self,
         operation: qasm3_ast.QuantumGate,
         inverse: bool = False,
--- a/src/pyqasm/visitor.py
+++ b/src/pyqasm/visitor.py
@@ -1698,10 +1698,6 @@ class QasmVisitor:
                 )
 
             assert isinstance(rhs_value, (bool, int))
-            # get classical registers
-            if reg_name is not None and reg_idx is not None and self._in_branching_statement:
-                op_tuple = (reg_name, reg_idx)
-                self._is_branch_clbits.add(op_tuple)
 
             if_block = self.visit_basic_block(statement.if_block)
             else_block = self.visit_basic_block(statement.else_block)
@@ -1711,6 +1707,9 @@ class QasmVisitor:
                 Qasm3Validator.validate_register_index(
                     reg_idx, self._global_creg_size_map[reg_name], qubit=False, op_node=condition
                 )
+
+                # getting creg for depth counting
+                self._is_branch_clbits.add((reg_name, reg_idx))
 
                 new_if_block = qasm3_ast.BranchingStatement(
                     condition=qasm3_ast.BinaryExpression(
@@ -1743,6 +1742,8 @@ class QasmVisitor:
                     rhs_value -= 1
 
                 size = self._global_creg_size_map[reg_name]
+                # getting cregs for depth counting
+                self._is_branch_clbits.update((reg_name, i) for i in range(size))
                 rhs_value_str = bin(int(rhs_value))[2:].zfill(size)
                 else_block = self.visit_basic_block(statement.else_block)
 
--- a/src/pyqasm/visitor.py
+++ b/src/pyqasm/visitor.py
@@ -859,7 +859,7 @@ class QasmVisitor:
             if not self._is_branching_statement: # if gate is not in branching statement
                 self._update_qubit_depth_for_gate(unrolled_targets, ctrls)
             else:
-                for ops in unrolled_targets + ctrls : # get qubit registers in branching operations
+                for ops in unrolled_targets + [ctrls] : # get qubit registers in branching operations
                     for op in ops:
                         op_idx = Qasm3ExprEvaluator.evaluate_expression(op.indices[0][0])[0]
                         op.name.name
@@ -967,7 +967,7 @@ class QasmVisitor:
             if not self._is_branching_statement: # if custom gate is not in branching statement
                 self._update_qubit_depth_for_gate([op_qubits], ctrls)
             else:
-                for ops in [op_qubits] + ctrls: # get qubit registers in branching operations
+                for ops in [op_qubits] + [ctrls]: # get qubit registers in branching operations
                     for op in ops:
                         op_idx = Qasm3ExprEvaluator.evaluate_expression(op.indices[0][0])[0]
                         op.name.name
--- a/src/pyqasm/visitor.py
+++ b/src/pyqasm/visitor.py
@@ -1650,9 +1650,8 @@ class QasmVisitor:
         for node in nodes:
             node.depth = max_depths
 
-        if not self._is_branching_statement:
-            self._is_branch_clbits.clear()
-            self._is_branch_qubits.clear()
+        self._is_branch_clbits.clear()
+        self._is_branch_qubits.clear()
 
     def _visit_branching_statement(
         self, statement: qasm3_ast.BranchingStatement
@@ -1781,8 +1780,7 @@ class QasmVisitor:
         self._curr_scope -= 1
         self._pop_scope()
         self._restore_context()
-        if self._is_branching_statement:
-            self._update_branching_gate_depths()
+        self._update_branching_gate_depths()
 
         if self._check_only:
             return []
--- a/src/pyqasm/visitor.py
+++ b/src/pyqasm/visitor.py
@@ -502,7 +502,7 @@ class QasmVisitor:
                         measure=qasm3_ast.QuantumMeasurement(qubit=src_id), target=None
                     )
                 )
-                if not self._is_branching_statement: # if measurement gate is not in branching statement
+                if not self._in_branching_statement: # if measurement gate is not in branching statement
                     src_name, src_id = src_id.name.name, src_id.indices[0][0].value  # type: ignore
                     qubit_node = self._module._qubit_depths[(src_name, src_id)]
                     qubit_node.depth += 1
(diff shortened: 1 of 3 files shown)


src/pyqasm/visitor.py
```diff
@@ -89,7 +89,9 @@ def __init__(
         self._curr_scope: int = 0
         self._label_scope_level: dict[int, set] = {self._curr_scope: set()}
         self._recording_ext_gate_depth = False
-
+        self._is_branching_statement: bool = False
```
TheGupta2012, line 92:
```suggestion
        self._in_branching_statement: bool = False
```

Let's rename this as we want to signify that we are inside a branching statement

src/pyqasm/visitor.py
```diff
@@ -1610,6 +1630,29 @@ def _evaluate_array_initialization(
...
+        for cbit in self._is_branch_clbits:
+            c_name, c_idx = cbit
+            c_node = self._module._clbit_depths[(c_name, c_idx)]
+            nodes.append(c_node)
+            max_depths = max(max_depths,c_node.depth + 1)
+        
+        for node in nodes:
+            node.depth = max_depths
+
+        if not self._is_branching_statement:
```
TheGupta2012, line 1653: Wouldn't this always be executed as we make it `False` at the start?

TheGupta2012 (changes requested):
Hey @vinayswamik , thanks for working on this! Giving a preliminary review right now, but the code looks good! 

Please add some more tests whenever you get time

src/pyqasm/visitor.py
vinayswamik: Yeah, That's true. i will do the changes.

src/pyqasm/visitor.py
```diff
@@ -1734,6 +1781,8 @@ def ravel(bit_ind):
         self._curr_scope -= 1
         self._pop_scope()
         self._restore_context()
+        if self._is_branching_statement:
```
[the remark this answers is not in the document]
vinayswamik: yes, i will update this too.

src/pyqasm/visitor.py
```diff
@@ -952,7 +968,16 @@ def _visit_custom_gate_operation(
         # Update the depth only once for the entire custom gate
         if self._recording_ext_gate_depth:
             self._recording_ext_gate_depth = False
-            self._update_qubit_depth_for_gate([op_qubits], ctrls)
+            if not self._in_branching_statement:  # if custom gate is not in branching statement
+                self._update_qubit_depth_for_gate([op_qubits], ctrls)
+            else:
+                for ops in [op_qubits] + [ctrls]:  # get qubit registers in branching operations
```
TheGupta2012, line 974: `op` and `ops` is a little bit confusing given the fact that we are referring to qubit registers here. Can we change this (and [here](https://github.com/qBraid/pyqasm/pull/200/files#diff-802ca57af29fc92ca8eb2f536fb5217c61c874c4c35bd39b7fd82512ee0e9a84R865)) to `qubit_subset` and `qubit` instead?

src/pyqasm/visitor.py
```diff
@@ -1649,6 +1698,10 @@ def _visit_branching_statement(
                 )
 
             assert isinstance(rhs_value, (bool, int))
+            # get classical registers
```
TheGupta2012, line 1701: - Should this not be present in line #1710 and #1733? Essentially, the check for `reg_idx` determines whether we have a case like `if ( c[1] == 1 )` OR `if ( c == 4 )`. 
- In the former case, your logic will work BUT if we are in the full register case, the classical bits **will not be added** to the `self._is_branch_clbits` set. You need to make sure that both these cases work out!

TheGupta2012: Hmm, I tested the case as well but it seems like the unrolled qasm is giving the correct depth - 

```python
In [32]: import pyqasm 

In [33]: qasm = """
    ...: OPENQASM 3;
    ...: include "stdgates.inc";
    ...: 
    ...: gate hgate q { h q; }
    ...: gate xgate q { x q; }
    ...: 
    ...: const int[32] N = 4;
    ...: qubit[4] q;
    ...: qubit ancilla;
    ...: 
    ...: crz(0.1) q[0], q[1];
    ...: 
    ...: def deutsch_jozsa(qubit[N] q_func, qubit[1] ancilla_q) {
    ...:   xgate ancilla_q;
    ...:   for int i in [0:N-1] { hgate q_func[i]; }
    ...:   hgate ancilla_q;
    ...:   for int i in [0:N-1] { cx q_func[i], ancilla_q; }
    ...:   for int i in [0:N-1] { hgate q_func[i]; }
    ...: }
    ...: 
    ...: deutsch_jozsa(q, ancilla);
    ...: 
    ...: bit[4] result;
    ...: result = measure q;
    ...: """

In [34]: module = pyqasm.loads(qasm)

In [35]: module.unroll()

In [36]: module.depth()
Out[36]: 8

In [37]: unrolled_qasm = pyqasm.dumps(module)

In [38]: unrolled_module = pyqasm.loads(unrolled_qasm)

In [39]: unrolled_module.depth()
Out[39]: 19
```

This is most likely a bug in the depth calculation with subroutines as the `unrolled_qasm` only has sequential instructions

src/pyqasm/visitor.py
TheGupta2012: How was this resolved?

src/pyqasm/visitor.py
vinayswamik: Thanks for the clarification 👍 . I will work on the rest.

vinayswamik:
> Changes lgtm @vinayswamik ! Can you please also add a `CHANGELOG.md` entry ? It will be good if you can specify an example program there. Post that, we can merge.
> 
> Thanks for working on this!

Thanks for your support @TheGupta2012. I have update the `CHANGELOG.md` and added one new test case to cover external gate depth inside branching statements. Let me know if any changes are needed.

TheGupta2012 (approved):
Good to merge

PR #205, opened 2025-06-11
Feature/Added explicit casting support

### LINKED ISSUE #10: [FEATURE] Add support for explicit casting
### Feature Description

Users can use explicit casting format to cast values from 1 type to another. Eg. - 
```c
const float[64] f1 = 2.5;
uint[8] runtime_u = 7;

// Valid statements

const int[8] i1 = int[8](f1);  // `i1` has compile-time value 2.
const uint u1 = 2 * uint(f1);  // `u1` has compile-time value 4.

// Invalid statements

const bit[2] b1 = bit[2](f1);  // `float[64]` cannot be cast to `bit[2]`.
const int[16] i2 = int[16](runtime_u);  // Casting runtime values is not `const`.
```
The `cast` operator translates to the `Cast` object in `openqasm` parser but is not parsed by our semantic analyser. We should be able to handle explicit casting statements and have the ability to identify which statements are valid / invalid. 

### Implementation (Optional)

Most likely changes will be needed in the declaration and assignment handlers for the semantic analyser. Although we have support for implicit casting, we would need to expand to also consider the exact sizes mentioned in the explicit type cast.

vinayswamik:
## Summary of changes

This PR adds initial support for **Explicit Casting** in both `QasmVisitor` and `Qasm3ExprEvaluator`, in accordance with the [OpenQASM 3 casting rules](https://openqasm.com/language/types.html#casting-specifics).

With this change, we introduce logic to:

- Convert values by reusing the existing implicit‐casting mechanism, then return the correctly cast result.
- Validate declaration vs. explicit cast, ensuring that the target type and size exactly match the variable’s original declaration; otherwise emits a `ValidationError`.
- Extended the test suite to cover all newly legal explicit‐cast scenarios as well as illegal‐cast error cases. 

Closes #10

### DIFF
--- a/src/pyqasm/expressions.py
+++ b/src/pyqasm/expressions.py
@@ -18,8 +18,10 @@ Module containing the class for evaluating QASM expressions.
 """
 from openqasm3.ast import (
     BinaryExpression,
+    BitType,
     BooleanLiteral,
     BoolType,
+    Cast,
     DurationLiteral,
     Expression,
     FloatLiteral,
@@ -40,6 +42,7 @@ from openqasm3.ast import (
 )
 
 from pyqasm.analyzer import Qasm3Analyzer
+from pyqasm.elements import Variable
 from pyqasm.exceptions import ValidationError, raise_qasm3_error
 from pyqasm.maps.expressions import CONSTANTS_MAP, qasm3_expression_op_map
 from pyqasm.validator import Qasm3Validator
@@ -49,6 +52,7 @@ class Qasm3ExprEvaluator:
     """Class for evaluating QASM3 expressions."""
 
     visitor_obj = None
+    cast_var_name = ""
 
     @classmethod
     def set_visitor_obj(cls, visitor_obj) -> None:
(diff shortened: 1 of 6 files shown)


src/pyqasm/visitor.py
```diff
@@ -462,6 +462,80 @@ def _get_op_bits(
...
+        base_size = 1
+        if not isinstance(base_type, qasm3_ast.BoolType):
+            initial_size = 1 if isinstance(base_type, qasm3_ast.BitType) else 32
+            try:
+                base_size = (
+                    initial_size
+                    if not hasattr(base_type, "size") or base_type.size is None
+                    else Qasm3ExprEvaluator.evaluate_expression(base_type.size, const_expr=True)[0]
+                )
+                if is_const:
```
TheGupta2012, line 487: Shouldn't this be executed in every case irrespective of whether we have const or not? Eg. for `int[-16] a;` this should be invoked as well right?

src/pyqasm/expressions.py
```diff
@@ -49,6 +52,7 @@ class Qasm3ExprEvaluator:
     """Class for evaluating QASM3 expressions."""
 
     visitor_obj = None
+    cast_var_name = ""
```
TheGupta2012, line 55: Could you elaborate as to why a class level variable is required for the variable which is being cast?

src/pyqasm/visitor.py
```diff
@@ -462,6 +462,80 @@ def _get_op_bits(
 
         return openqasm_bits
 
+    def _check_type_size(
```
TheGupta2012, line 465: This can be moved to the [`Qasm3Validator`](https://github.com/qBraid/pyqasm/blob/main/src/pyqasm/validator.py#L41) class as it has minimal dependency on the visitor and is only validating the size

src/pyqasm/visitor.py
```diff
@@ -462,6 +462,80 @@ def _get_op_bits(
...
+                raise_qasm3_error(
+                    f"Invalid base size for {err_str} '{var_name}'",
+                    error_node=statement,
+                    span=statement.span,
+                    raised_from=err,
+                )
+        return base_size
+
+    # pylint: disable-next=too-many-arguments
+    def _check_type_for_cast(
```
TheGupta2012, line 505: Similar comment as above, can be moved to the validator class

TheGupta2012 (changes requested):
Hi @vinayswamik , thank you for the great work! I have left some comments regarding organization of the code but overall it looks good. 

Can you please also add a note in the CHANGELOG and an entry for the `Cast` in [pyqasm supported instructions](https://github.com/qBraid/pyqasm/tree/main/src#pyqasm) ?

src/pyqasm/expressions.py
vinayswamik: I changed this to local variable. we don't need class level attribute. 👍

src/pyqasm/visitor.py
vinayswamik: we can do that, but the problem is **circular import**. 
because, `_check_type_size` is in `expressions.py`, which is trying to import `Qasm3Validator` from `validator.py`, But `validator.py` is also importing from `expressions.py` before it's fully loaded. 

one simple way is, local import inside the function. can i do that..?

TheGupta2012:
Hi @vinayswamik , any reasons why you closed the PR?

src/pyqasm/visitor.py
TheGupta2012: Local imports aren't really good practice. Usually it is possible to find a way to re-factor code to avoid circular imports.

vinayswamik: yeah, instead we can move them to `expressions.py` as a static methods. can i do that...?

TheGupta2012 (approved):
Thank you @vinayswamik for the changes, LGTM!

PR #210, opened 2025-06-18
Fix: Error in circuit visualization of decomposable gates

### LINKED ISSUE #209: [BUG] Error in circuit visualization with decomposable gates
### Environment

- **PyQASM version**: 0.4.0
- **Python version**: 3.13.2
- **Operating system**: MacOS 15.5

### What happened?

pyqasm `draw` method raises error for circuits with decomposable gates.

```python
import pyqasm

qasm = """
OPENQASM 3.0;
qubit[2] q1;
qreg q[3];
creg c[3];
ccx q[0], q[1], q1[0];
crx (0.1) q[0], q[1];
rccx q[0], q[1], q1[0];
"""

qc = pyqasm.loads(qasm)
qc.validate()
pyqasm.draw(qc)
```
```python
Error: AttributeError: 'numpy.ndarray' object has no attribute 'set_ylim'
```

### Suggestions (Optional)

When we have decomposable gates, the circuit visualization code is creating multiple subplots to show the decomposed operations. In these cases, matplotlib's `plt.subplots()` is returning a NumPy array of axes instead of a list of matplotlib Axes objects. Our code was trying to call `set_ylim()` on this NumPy array instead of on individual Axes objects.
We need to ensure that we're always working with a list of matplotlib Axes objects, regardless of whether we have one or multiple subplots.

vinayswamik:
Fixes #209
## Summary of changes
- This fix makes axes handling more robust by first checking whether axis is a numpy array and, if so, converting it into a flat list of Axes objects.
- If axis is a single instance, it’s wrapped in a list to ensure consistency. Finally, if axis is already a list of Axes, it’s left unchanged. 
- This guarantees that downstream code always works with a proper list of Axes objects, regardless of the number of subplots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ Types of changes:
 ### Removed
 
 ### Fixed
-
+- Fixed multiple axes error in circuit visualization of decomposable gates in `draw` method. ([#209](https://github.com/qBraid/pyqasm/pull/210))
 ### Dependencies
 
 ### Other
--- a/src/pyqasm/printer.py
+++ b/src/pyqasm/printer.py
@@ -297,6 +297,7 @@ def _mpl_setup_figure(
     sections: list[list[list[QuantumStatement]]], width: float, n_lines: int
 ) -> tuple[plt.Figure, list[plt.Axes]]:
     import matplotlib.pyplot as plt
+    import numpy as np
 
     fig_ax_tuple: tuple[plt.Figure, list[plt.Axes] | plt.Axes] = plt.subplots(
         len(sections),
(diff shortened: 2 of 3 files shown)


vinayswamik:
> Thanks for fixing this @vinayswamik ! While testing out the code I see that the gates are unrolled by default -
> 
> ```python
> from pyqasm import loads
> from pyqasm.printer import mpl_draw
> 
> qasm = """
>     OPENQASM 3.0;
>     qubit[2] q1;
>     qreg q[3];
>     creg c[3];
>     ccx q[0], q[1], q1[0];
>     crx (0.1) q[0], q[1];
>     rccx q[0], q[1], q1[0];
>     cz q[0], q[1];
>     """
> circ = loads(qasm)
> mpl_draw(circ, filename = 'test.png')
> ```
> 
> image
> 
> Can you see why this is happening? Ideally we should represent the gates as is, unless an `unroll` is called internally

The reason is `program.unroll()` is calling in the beginning of `mpl_draw` function.

TheGupta2012:
Hmm, ideally it should be `validate` as we don't wanna change the representation unless the user triggers an `unroll` command.

PR #211, opened 2025-06-18
Fix: depth calculations for decompose gates

### LINKED ISSUE #201: [BUG] Depth mismatch with QASM3 subroutines
### Environment

- **PyQASM version**: 0.3.2
- **Python version**: 3.12.1
- **Operating system**: Ubuntu 20.04.6 LTS

### What happened?

Depth calculation for the following program is inconsistent - 

```python
In [32]: import pyqasm 

In [33]: qasm = """
    ...: OPENQASM 3;
    ...: include "stdgates.inc";
    ...: 
    ...: gate hgate q { h q; }
    ...: gate xgate q { x q; }
    ...: 
    ...: const int[32] N = 4;
    ...: qubit[4] q;
    ...: qubit ancilla;
    ...: 
    ...: crz(0.1) q[0], q[1];
    ...: 
    ...: def deutsch_jozsa(qubit[N] q_func, qubit[1] ancilla_q) {
    ...:   xgate ancilla_q;
    ...:   for int i in [0:N-1] { hgate q_func[i]; }
    ...:   hgate ancilla_q;
    ...:   for int i in [0:N-1] { cx q_func[i], ancilla_q; }
    ...:   for int i in [0:N-1] { hgate q_func[i]; }
    ...: }
    ...: 
    ...: deutsch_jozsa(q, ancilla);
    ...: 
    ...: bit[4] result;
    ...: result = measure q;
    ...: """

In [34]: module = pyqasm.loads(qasm)

In [35]: module.unroll()

In [36]: module.depth()
Out[36]: 8

In [37]: unrolled_qasm = pyqasm.dumps(module)

In [38]: unrolled_module = pyqasm.loads(unrolled_qasm)

In [39]: unrolled_module.depth()
Out[39]: 19
```

### Suggestions (Optional)

The depth calculation should be exactly the same for the module before and after unrolling. The culprit seems like the subroutine application but needs some investigation

vinayswamik:
Fixes #201
#### Added special handling for depth calculation for decompose gates
- A new property `decompose_gates` was added to the `QasmModule` class to control whether gates should be decomposed during depth calculations.
- This allows for more flexible depth calculations where users can choose whether to consider decomposed gates or not.

### DIFF
--- a/src/pyqasm/modules/base.py
+++ b/src/pyqasm/modules/base.py
@@ -56,7 +56,7 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
         self._validated_program = False
         self._unrolled_ast = Program(statements=[])
         self._external_gates: list[str] = []
-        self._decompose_gates: Optional[bool] = None
+        self._decompose_native_gates: Optional[bool] = None
 
     @property
     def name(self) -> str:
@@ -282,7 +282,7 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
         qasm_module = self.copy()
         qasm_module._qubit_depths = {}
         qasm_module._clbit_depths = {}
-        qasm_module._decompose_gates = decompose_gates
+        qasm_module._decompose_native_gates = decompose_gates
         # Unroll using any external gates that have been recorded for this
         # module
         qasm_module.unroll(external_gates=self._external_gates)
--- a/src/pyqasm/modules/base.py
+++ b/src/pyqasm/modules/base.py
@@ -260,11 +260,11 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
 
         return curr_module
 
-    def depth(self, decompose_gates=True):
+    def depth(self, decompose_native_gates=True):
         """Calculate the depth of the unrolled openqasm program.
 
         Args:
-        decompose_gates (bool): If True, calculate depth after decomposing gates.
+        decompose_native_gates (bool): If True, calculate depth after decomposing gates.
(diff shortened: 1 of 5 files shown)


tests/qasm3/test_depth.py
```diff
@@ -624,3 +641,26 @@ def test_qasm3_depth_branching_for_external_gates():
...
+"""
+
+
+@pytest.mark.parametrize(
+    ["input_qasm_str", "before_decompose", "after_decompose"],
+    [
+        (QASM3_DECOMPOSE_GATE_DEPTH_1, 2, 25),
+    ],
+)
+def test_gate_depth_decomposable_gates(input_qasm_str, before_decompose, after_decompose):
```
TheGupta2012, line 662: Can also add a test where we have custom gates (which should be unrolled), external gate (which should be added as depth = 1) and then this new parameter in `depth`

src/pyqasm/modules/base.py
```diff
@@ -259,11 +260,13 @@ def remove_includes(self, in_place=True) -> Optional["QasmModule"]:
 
         return curr_module
 
-    def depth(self):
+    def depth(self, decompose_gates=True):
```
TheGupta2012, line 263:
```suggestion
    def depth(self, decompose_native_gates=True):
```

TheGupta2012: Could you rename this as well? The parameter is user facing and it is better to specify the same name as the one used in the `QasmModule`

TheGupta2012 (approved):
LGTM, thanks!

PR #222, opened 2025-06-27
Feature/Implement Qubit Register Consolidation

### LINKED ISSUE #187: [FEATURE] Qubit Register Consolidation
## **Feature Description**
- Introduce a `device_qubits` parameter to PyQASM’s `unroll` method that consolidates all declared qubit and qreg registers in a QASM program into a single, uniquely named register: `qubit[N] __PYQASM_QUBITS__`, where `N` is the total number of qubits specified by the user. 
- When `device_qubits` is specified, `__PYQASM_QUBITS__` becomes a special reserved keyword and cannot be used elsewhere in the user’s code. All gate and measurement operations referencing original registers are remapped to this consolidated register. If the total number of required qubits exceeds the specified device size, an error is raised.

**The order in which qubit registers appear in the original QASM program defines their mapping in the consolidated register:** the first declared register occupies the lowest indices, followed by the next, and so on. This ensures deterministic and predictable assignment of qubit IDs within `__PYQASM_QUBITS__`.

## **Motivation**
- **Hardware Compatibility:** Many quantum devices require flat, contiguous qubit addressing.
- **Simplified Compilation:** Automates the tedious and error-prone process of remapping multiple logical registers to a single hardware register.
- **Error Prevention:** Validates that the user’s program does not exceed the available device qubits, preventing invalid execution.
- **Consistent Register Management:** Maintaining a single, reserved register (`__PYQASM_QUBITS__`) throughout the program simplifies code generation, post-processing, and downstream analysis.
- **Facilitates Routing:** Consolidating all qubits into one register makes routing and mapping phases more straightforward.

## **Implementation**
- **Input Handling:**  
  - Add `device_qubits: int` as an optional parameter to `unroll()`.
  - Compute the total number of qubits required by summing all register and qubit declarations.
- **Validation:**  
  - Raise an error if the total required qubits exceed `device_qubits`.
  - Raise an error if the original QASM program declares a register named `__PYQASM_QUBITS__` when `device_qubits` is specified.
- **Consolidation:**  
  - Remove and replace all qubit and qreg declarations with `qubit[device_qubits] __PYQASM_QUBITS__;`.
  - Rewrite all gate and measurement operands to reference `__PYQASM_QUBITS__[i]`, where `i` is the offset determined by the order of appearance of each register in the original code.
- **Edge Cases:**  
  - Ensure no overlap in qubit indexing and preserve original program semantics.

---

## Example Workflow
```python
from pyqasm import loads

qasm_code = """
OPENQASM 3.0;
qubit[4] data;
qubit[2] ancilla;
cx data[0], ancilla[1];
"""

module = loads(qasm_code)

# User specifies a 6-qubit device
module.unroll(device_qubits = 6)

print(module)

# Output:
# OPENQASM 3.0;
# qubit[6] __PYQASM_QUBITS__;
# cx __PYQASM_QUBITS__[0], __PYQASM_QUBITS__[5];
```
In this example, the data register (4 qubits) appears first and is mapped to `__PYQASM_QUBITS__[0-3]`, while the ancilla register (2 qubits) appears next and is mapped to `__PYQASM_QUBITS__[4-5]`.

> **NOTE**: When device_qubits is specified, `__PYQASM_QUBITS__` is a reserved keyword. Any attempt to declare or use `__PYQASM_QUBITS__` as a register name in the original QASM program will result in an error. The mapping of original registers to the consolidated register strictly follows their order of appearance in the source code.

vinayswamik:
## Summary of changes
Implement device qubit(`__PYQASM_QUBITS__`) support in QasmVisitor and QasmModule

- Added a new parameter `device_qubits` to `QasmVisitor` and `QasmModule` to manage qubit consolidation.
- Introduced `_get_pyqasm_device_qubit_index`  method for qubit index mapping and `_qubit_register_consolidation` to ensure compatibility with device constraints.
- Updated measurement, barrier, and gate operations to utilize the new device qubit logic.
- Added unit tests to validate the functionality of qubit consolidation and error handling for exceeding device qubit limits.
enhance a little bit wihtout changing anuthing

Closes #187

### DIFF
--- a/src/pyqasm/entrypoint.py
+++ b/src/pyqasm/entrypoint.py
@@ -30,11 +30,15 @@ if TYPE_CHECKING:
     import openqasm3.ast
 
 
-def load(filename: str) -> QasmModule:
+def load(
+    filename: str, *, device_qubits: int | None = None, consolidate_qubits: bool = False
+) -> QasmModule:
     """Loads an OpenQASM program into a `QasmModule` object.
 
     Args:
         filename (str): The filename of the OpenQASM program to validate.
+        device_qubits (int): Number of physical qubits available on the target device.
+        consolidate_qubits (bool): If True, consolidate all quantum registers into single register.
 
     Returns:
         QasmModule: An object containing the parsed qasm representation along with
@@ -44,14 +48,21 @@ def load(filename: str) -> QasmModule:
         raise TypeError("Input 'filename' must be of type 'str'.")
     with open(filename, "r", encoding="utf-8") as file:
         program = file.read()
-    return loads(program)
+    return loads(program, device_qubits=device_qubits, consolidate_qubits=consolidate_qubits)
 
 
-def loads(program: openqasm3.ast.Program | str) -> QasmModule:
+def loads(
+    program: "openqasm3.ast.Program | str",
+    *,
+    device_qubits: int | None = None,
+    consolidate_qubits: bool = False,
+) -> QasmModule:
     """Loads an OpenQASM program into a `QasmModule` object.
 
     Args:
         program (openqasm3.ast.Program or str): The OpenQASM program to validate.
+        device_qubits (int): Number of physical qubits available on the target device.
+        consolidate_qubits (bool): If True, consolidate all quantum registers into single register.
 
     Raises:
         TypeError: If the input is not a string or an `openqasm3.ast.Program` instance.
@@ -79,7 +90,9 @@ def loads(program: openqasm3.ast.Program | str) -> QasmModule:
 
     qasm_module = Qasm3Module if program.version.startswith("3") else Qasm2Module
     module = qasm_module("main", program)
-
+    # Store device_qubits and consolidate_qubits on the module for later use
+    module._device_qubits = device_qubits
+    module._consolidate_qubits = consolidate_qubits
     return module
 
 
--- a/src/pyqasm/entrypoint.py
+++ b/src/pyqasm/entrypoint.py
@@ -31,7 +31,7 @@ if TYPE_CHECKING:
 
 
 def load(
-    filename: str, *, device_qubits: int | None = None, consolidate_qubits: bool = False
+    filename: str, device_qubits: int | None = None, consolidate_qubits: bool = False
 ) -> QasmModule:
     """Loads an OpenQASM program into a `QasmModule` object.
 
@@ -52,8 +52,7 @@ def load(
 
 
 def loads(
-    program: "openqasm3.ast.Program | str",
-    *,
+    program: openqasm3.ast.Program | str,
     device_qubits: int | None = None,
     consolidate_qubits: bool = False,
 ) -> QasmModule:
--- a/src/pyqasm/entrypoint.py
+++ b/src/pyqasm/entrypoint.py
@@ -30,15 +30,11 @@ if TYPE_CHECKING:
     import openqasm3.ast
 
 
-def load(
-    filename: str, device_qubits: int | None = None, consolidate_qubits: bool = False
-) -> QasmModule:
+def load(filename: str, **kwargs) -> QasmModule:
     """Loads an OpenQASM program into a `QasmModule` object.
 
     Args:
         filename (str): The filename of the OpenQASM program to validate.
-        device_qubits (int): Number of physical qubits available on the target device.
-        consolidate_qubits (bool): If True, consolidate all quantum registers into single register.
 
     Returns:
         QasmModule: An object containing the parsed qasm representation along with
@@ -48,20 +44,17 @@ def load(
         raise TypeError("Input 'filename' must be of type 'str'.")
     with open(filename, "r", encoding="utf-8") as file:
         program = file.read()
-    return loads(program, device_qubits=device_qubits, consolidate_qubits=consolidate_qubits)
+    return loads(program, **kwargs)
 
 
-def loads(
-    program: openqasm3.ast.Program | str,
-    device_qubits: int | None = None,
-    consolidate_qubits: bool = False,
-) -> QasmModule:
+def loads(program: openqasm3.ast.Program | str, **kwargs) -> QasmModule:
     """Loads an OpenQASM program into a `QasmModule` object.
 
     Args:
         program (openqasm3.ast.Program or str): The OpenQASM program to validate.
+
+    **kwargs: Additional arguments to pass to the loads function.
         device_qubits (int): Number of physical qubits available on the target device.
-        consolidate_qubits (bool): If True, consolidate all quantum registers into single register.
 
     Raises:
         TypeError: If the input is not a string or an `openqasm3.ast.Program` instance.
@@ -89,9 +82,9 @@ def loads(
 
     qasm_module = Qasm3Module if program.version.startswith("3") else Qasm2Module
     module = qasm_module("main", program)
-    # Store device_qubits and consolidate_qubits on the module for later use
-    module._device_qubits = device_qubits
-    module._consolidate_qubits = consolidate_qubits
+    # Store device_qubits on the module for later use
+    if dev_qbts := kwargs.get("device_qubits"):
+        module._device_qubits = dev_qbts
     return module
 
 
--- a/src/pyqasm/entrypoint.py
+++ b/src/pyqasm/entrypoint.py
@@ -53,8 +53,8 @@ def loads(program: openqasm3.ast.Program | str, **kwargs) -> QasmModule:
     Args:
         program (openqasm3.ast.Program or str): The OpenQASM program to validate.
 
-    **kwargs: Additional arguments to pass to the loads function.
-        device_qubits (int): Number of physical qubits available on the target device.
+        **kwargs: Additional arguments to pass to the loads function.
+            device_qubits (int): Number of physical qubits available on the target device.
 
     Raises:
         TypeError: If the input is not a string or an `openqasm3.ast.Program` instance.
--- a/src/pyqasm/modules/base.py
+++ b/src/pyqasm/modules/base.py
@@ -57,6 +57,7 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
         self._unrolled_ast = Program(statements=[])
         self._external_gates: list[str] = []
         self._decompose_native_gates: Optional[bool] = None
+        self._device_qubits: Optional[int] = 0
 
     @property
     def name(self) -> str:
@@ -551,6 +552,8 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
                 self._external_gates = ext_gates
             else:
                 self._external_gates = []
+            if device_qbts := kwargs.get("device_qubits"):
+                self._device_qubits = device_qbts
             visitor = QasmVisitor(module=self, **kwargs)
             self.accept(visitor)
         except (ValidationError, UnrollError) as err:
--- a/src/pyqasm/modules/base.py
+++ b/src/pyqasm/modules/base.py
@@ -519,19 +519,14 @@ class QasmModule(ABC):  # pylint: disable=too-many-instance-attributes
             return
         try:
             self.num_qubits, self.num_clbits = 0, 0
-            visitor = QasmVisitor(
-                self,
-                check_only=True,
-                device_qubits=self._device_qubits,
-            )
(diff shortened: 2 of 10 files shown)


src/pyqasm/modules/base.py
```diff
@@ -57,6 +57,7 @@ def __init__(self, name: str, program: Program):
         self._unrolled_ast = Program(statements=[])
         self._external_gates: list[str] = []
         self._decompose_native_gates: Optional[bool] = None
+        self._device_qubits: Optional[int] = 0
```
TheGupta2012, line 60:
```suggestion
        self._device_qubits: Optional[int] = None
```

TheGupta2012: Also, let's make this a `kwarg` of the `load(s)` method, eg. - 

```python
module = pyqasm.loads(qasm, device_qubits = 23)
```

The idea is that a module will be defined using a specific number of device qubits but the users can "choose to consolidate" the qasm according to the device specification. Then, the corresponding argument in `unroll` becomes `consolidate_qubits` or equivalent. 

**NOTE**: this will also help us implicitly validate the total qubit count according to the total qubits on a device. We can add this checking inside the `validate` functionality

src/pyqasm/visitor.py
```diff
@@ -86,6 +86,7 @@ def __init__(  # pylint: disable=too-many-arguments
         external_gates: list[str] | None = None,
         unroll_barriers: bool = True,
         max_loop_iters: int = int(1e9),
+        device_qubits: int = 0,
```
TheGupta2012, line 89: Derive from `module`, use something like a `_consolidate_qubits` flag to check and perform the consolidation

src/pyqasm/visitor.py
```diff
@@ -559,7 +562,66 @@ def _check_variable_cast_type(
...
+        if "__PYQASM_QUBITS__" in self._global_qreg_size_map:
+            raise_qasm3_error(
+                "Original QASM program already declares reserved register '__PYQASM_QUBITS__'."
+            )
+
+        removable_statements = []
+        for stmt in unrolled_stmts:
+            if isinstance(stmt, qasm3_ast.QubitDeclaration):
+                removable_statements.append(stmt)
+        for r_stmt in removable_statements:
```
TheGupta2012, line 616: It is better to construct a new list of `valid_statements` which do not contain the original `QubitDeclaration`s. You can just append any statement which is not of type `QubitDeclaration`. 

The reason is that `.remove(...)` method works in `O(n)` time, `n` being the size of our list. For a large number of qubit declarations, say `k`, this will scale as `O(n*k)`.

TheGupta2012 (changes requested):
Hi @vinayswamik , the changes look good! 

A point to consider for the design - 

Consolidation of qubits inside the qasm operations should be done as a "finalize" step. Presently, you are updating the qubit identifiers directly during the unroll / validate phase and subsequently adding the `__PYQASM_QUBITS__` declaration. While it works for the test cases, this transformation logic should be separated in a new step. 

Building upon this logic, the transformation of statements which you have done inside each of `reset`, `barrier`, etc. should be done as part of the `_qubit_register_consolidation` method. It should read the unrolled statements and transform the qubits for each of them, possibly utilizing the `Qasm3Transformer` module. I majorly see `self._get_pyqasm_device_qubit_index` being utilized in your changes and can be decoupled from the visitor code.

src/pyqasm/visitor.py
```diff
@@ -559,7 +562,66 @@ def _check_variable_cast_type(
...
+            unrolled_stmts.remove(r_stmt)
+        pyqasm_reg_id = qasm3_ast.Identifier("__PYQASM_QUBITS__")
+        pyqasm_reg_size = qasm3_ast.IntegerLiteral(self._device_qubits)
+        pyqasm_reg_stmt = qasm3_ast.QubitDeclaration(pyqasm_reg_id, pyqasm_reg_size)
+        unrolled_stmts.insert(1, pyqasm_reg_stmt)
```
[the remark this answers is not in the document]
vinayswamik: for any valid qasm3 code first stmt is include(stdgates) in unrolled statements. but there might be other include stmts in future. one better way would be below approach, which costs only O(n).

```python
pyqasm_reg_id = qasm3_ast.Identifier("__PYQASM_QUBITS__")
pyqasm_reg_size = qasm3_ast.IntegerLiteral(self._device_qubits)
pyqasm_reg_stmt = qasm3_ast.QubitDeclaration(pyqasm_reg_id, pyqasm_reg_size)

 valid_statements = []
_inserted_pyqasm_reg = False
for stmt in unrolled_stmts:
      if not isinstance(stmt, qasm3_ast.QubitDeclaration):
          continue
      if not _inserted_pyqasm_reg and not isinstance(stmt, qasm3_ast.Include):
          valid_statements.append(pyqasm_reg_stmt)
          _inserted_pyqasm_reg = True
      valid_statements.append(stmt)
```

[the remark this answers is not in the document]
TheGupta2012: Well, we mostly see people adding `include` at the top of the programs but the following program is syntactically correct as well - 

```qasm 
OPENQASM 3;
qubit[3] q;
reset q;
include "stdgates.inc";
h q;
```
According to the above logic, this will add the declaration of `__PYQASM_QUBITS__` after its first usage in the `reset q` stmt! 

On second thoughts, I believe declaring the `__PYQASM_QUBITS__` as the first statement itself would be the best way forward.

src/pyqasm/entrypoint.py
```diff
@@ -30,11 +30,15 @@
     import openqasm3.ast
 
 
-def load(filename: str) -> QasmModule:
+def load(
+    filename: str, device_qubits: int | None = None, consolidate_qubits: bool = False
```
TheGupta2012, line 34: Please add these add `kwargs` and keep `consolidate_qubits` on the `unroll` level

src/pyqasm/modules/base.py
```diff
@@ -517,8 +519,20 @@ def validate(self):
...
-            visitor = QasmVisitor(self, check_only=True)
+            visitor = QasmVisitor(
+                self,
+                check_only=True,
+                device_qubits=self._device_qubits,
+            )
             self.accept(visitor)
+            # Implicit validation: check total qubits if device_qubits is set and not consolidating
+            if self._device_qubits:
+                total_qubits = sum(self._qubit_registers.values())
```
TheGupta2012, line 530: Use `self._num_qubits`

src/pyqasm/modules/base.py
```diff
@@ -545,12 +562,21 @@ def unroll(self, **kwargs):
         """
         if not kwargs:
             kwargs = {}
+        # Use module attributes if not overridden by kwargs
+        if "device_qubits" not in kwargs:
```
TheGupta2012, line 566: Let us keep it only as a module level argument, do not mix with `unroll`

src/pyqasm/visitor.py
```diff
@@ -559,7 +576,54 @@ def _check_variable_cast_type(
...
+
+        if "__PYQASM_QUBITS__" in self._global_creg_size_map:
+            raise_qasm3_error(
+                "Original QASM program already declares classical register '__PYQASM_QUBITS__'.",
+            )
+
+        global_scope = self._get_global_scope()
+        if "__PYQASM_QUBITS__" in global_scope:
+            raise_qasm3_error(
+                "Variable '__PYQASM_QUBITS__' is already exists",
```
TheGupta2012, line 611:
```suggestion
                "Variable '__PYQASM_QUBITS__' is already defined",
```

Please also use the `span` of the variable in the error reporting

src/pyqasm/visitor.py
```diff
@@ -559,7 +576,54 @@ def _check_variable_cast_type(
...
+            ValidationError: If the total number of qubits exceeds the available device qubits,
+                             or if the reserved register '__PYQASM_QUBITS__' is already declared
+                             in the original QASM program.
+        """
+        if total_qubits > self._device_qubits:  # type: ignore
+            raise_qasm3_error(
+                f"Total qubits '({total_qubits})' exceed device qubits '({self._device_qubits})'.",
+            )
+
+        if "__PYQASM_QUBITS__" in self._global_qreg_size_map:
```
TheGupta2012, line 598: Do not need checks for qubits and bits separately. The global scope will already contain variables pertaining to each of the registers. See ref - [[1](https://github.com/qBraid/pyqasm/blob/73c2167d3ea77aaa3706c776ca8208b0d88b298e/src/pyqasm/visitor.py#L339)] , [[2](https://github.com/qBraid/pyqasm/blob/73c2167d3ea77aaa3706c776ca8208b0d88b298e/src/pyqasm/visitor.py#L1545)]

src/pyqasm/visitor.py
```diff
@@ -352,6 +360,15 @@ def _visit_quantum_register(
 
         self._module._add_qubit_register(register_name, register_size)
 
+        # Inline: Update offsets after adding a new register if device_qubits is set
+        if self._consolidate_qubits:
+            offsets = {}
```
TheGupta2012, line 365: We don't need to construct this repeatedly right? Just get the current max offset, add it as the offset of the current qubit register and then increment the max offset by current register size

Also, we need to preserve the ordering of the qubits appearing in our quantum program - using an [ordered dict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) should help keep the order of the qubit declarations.

src/pyqasm/transformer.py
```diff
@@ -438,3 +440,142 @@ def get_type_string(variable: Variable) -> str:
...
+                    list, unrolled_stmts.qubits[0].indices
+                )  # type: ignore[union-attr]
+                for multi_ind in _qubit_ind:
+                    for ind in multi_ind:
+                        pyqasm_ind = _get_pyqasm_device_qubit_index(
+                            _qubit_str.name, ind.value, qubit_register_offsets, global_qreg_size_map
+                        )
+                        ind.value = pyqasm_ind
+                _qubit_str.name = "__PYQASM_QUBITS__"
+            return unrolled_stmts
```
TheGupta2012, line 505: can move this out of the checks, because you are essentially doing this for every case

src/pyqasm/visitor.py
vinayswamik: at this level, we need extra code around the visitor, currently `Variable()`don't have span element. i will add that and search around the global scope variables and raise the error if found. 
Does that works..?

src/pyqasm/transformer.py
```diff
@@ -438,3 +440,142 @@ def get_type_string(variable: Variable) -> str:
...
+                                ind_val.value,
+                                qubit_register_offsets,
+                                global_qreg_size_map,
+                            )
+                            ind_val.value = pyqasm_val
+                    _qubit_ind_id.name.name = "__PYQASM_QUBITS__"
+                return unrolled_stmts
+
+            if isinstance(unrolled_stmts[0], QuantumGate):
+                unrolled_copy = deepcopy(unrolled_stmts)
```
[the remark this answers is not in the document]
vinayswamik: I’ve only seen this happen with decomposed gates. After unrolling, when I update the qubit indices on one gate statement, it also changes the qubit indices on other gate statements. I suspect that—even post-unrolling—they’re all sharing the same qubit-index identifier object.

That makes it difficult to read and assign in-place.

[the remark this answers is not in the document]
TheGupta2012: Yep, after `unroll` any ranges or full qubit registers are unrolled to single qubit indices

src/pyqasm/visitor.py
```diff
@@ -352,6 +366,17 @@ def _visit_quantum_register(
 
         self._module._add_qubit_register(register_name, register_size)
 
+        # Inline: Update offsets after adding a new register if device_qubits is set
+        if self._consolidate_qubits:
+            self._qubit_register_offsets[register_name] = self._qubit_register_max_offset
+            self._qubit_register_max_offset += register_size
+            # offsets = {}
```
TheGupta2012, line 373: You can remove the old code and just add a explanation as to why `self._qubit_register_offsets` is needed

TheGupta2012:
Besides the minor comment fix, changes lgtm @vinayswamik ! Once the CI fix is merged, can you rebase as well? Should be good to merge post that

TheGupta2012 (approved):
Lgtm

TheGupta2012 (approved):
Lgtm

PR #223, opened 2025-06-29
Fix/ Statement copying in functions and added loop limit in for loops

### LINKED ISSUE #202: Evaluate and replace `copy.deepcopy` with `copy.copy`
In the current function call visitor, we use [`deepcopy` to copy the function statements](https://github.com/qBraid/pyqasm/blob/7076d46e057085fbb6189dbe52d6e6ab4b2b2170/src/pyqasm/visitor.py#L1916) and visit them sequentially. If the statement internals are not modified, we can use shallow copy or `copy.copy` method which will improve the runtime for unrolling subroutines.

### LINKED ISSUE #208: [FEATURE] Add max_loop iteration limit in `_visit_forin_loop`
### Feature Description

With the introduction of `max_loop_iters` parameter in #206 , we want to add the max iteration check in unrolling for loops as well. 

### Implementation (Optional)

Simply check if the [`irange` size](https://github.com/qBraid/pyqasm/blob/c27b69e4c32823dfc83fe68dd206dbce616f4709/src/pyqasm/visitor.py#L1920) is going above the `max_loop_iters` or not. If yes, we break out.

vinayswamik:
#### Optimize statement copying in `_visit_function_call` and added `max_loop_iters` in `_visit_forin_loop`

- Optimized the handling of return statements and function operations by using shallow copies where appropriate, while maintaining fallback to deep copies in case of errors.

Closes #202
- Added a check to ensure loop ranges do not exceed the maximum allowed iterations, raising a LoopLimitExceededError if exceeded.

Closes #208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Types of changes:
 ### Fixed
 - Fixed multiple axes error in circuit visualization of decomposable gates in `draw` method. ([#209](https://github.com/qBraid/pyqasm/pull/210))
 - Fixed depth calculation for decomposable gates by computing depth of each constituent quantum gate.([#211](https://github.com/qBraid/pyqasm/pull/211))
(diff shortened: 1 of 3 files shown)


TheGupta2012 (approved):
LGTM

PR #229, opened 2025-07-10
Add OpenPulseVisitor Skeleton and Initial OpenPulse AST Handling

### LINKED ISSUE #183: [FEATURE] Use `openpulse.parser.parse_openpulse` to parse `cal` and `defcal` blocks in program
### Feature Description

As mentioned in #95 , the `openqasm3` parser does not parse the body of the calibration statements in its native parser. For each of the [`CalibrationDefinition`](https://github.com/openqasm/openqasm/blob/9150b157df801702b02ac7c14979ace8aba53a85/source/openqasm/openqasm3/ast.py#L861) and [`CalibrationStatement`](https://github.com/openqasm/openqasm/blob/9150b157df801702b02ac7c14979ace8aba53a85/source/openqasm/openqasm3/ast.py#L847) we need to be able to parse its body.

### Implementation (Optional)

If the `pulse` extra is installed, use the [`parse_openpulse`](https://github.com/openqasm/openpulse-python/blob/main/source/openpulse/openpulse/parser.py#L77) function in `openpulse` to correctly populate the body of the calibration statements.

vinayswamik:
## Summary of changes

This PR introduces the initial implementation of the `OpenPulseVisitor`. This visitor is designed to traverse and interact with elements in an `OpenPulse` program, providing a foundation for OpenPulse AST handling within the pyqasm framework.
This foundational work sets up the structure for `OpenPulse` support in pyqasm, enabling future development of OpenPulse parsing, validation, and transformation features. The current implementation provides scaffolding and type checks, with detailed TODOs for each OpenPulse construct.
- Implement the detailed logic for each visitor method (frame/port/waveform management, function call evaluation, etc.).
- Integrate with the rest of the pyqasm pipeline for end-to-end OpenPulse support.

Fixes #183

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Install dependencies
       run: |
         python3 -m pip install --upgrade pip
-        python3 -m pip install -e '.[docs]'
+        python3 -m pip install -e '.[docs,pulse]'
     - name: Build docs
       run: |
         sphinx-build -W -b html docs docs/build/html
--- a/src/pyqasm/elements.py
+++ b/src/pyqasm/elements.py
@@ -106,6 +106,25 @@ class Variable:  # pylint: disable=too-many-instance-attributes
     readonly: bool = False
 
 
+@dataclass(slots=True)
+class Frame:
+    """
+    Class representing a frame in OpenQASM/OpenPulse.
+
+    Attributes:
+        port (Any): The port to which the frame is attached. Immutable after initialization.
+        frequency (float): The frequency of the frame.
+        phase (float): The phase of the frame, in radians.
+        time (float): The current time of the frame, in duration units.
+                      This is manipulated implicitly and cannot be set directly.
+    """
+
+    port: Any
+    frequency: Any
+    phase: Any
+    time: Optional[Any] = 0.0
+
+
 class BasisSet(Enum):
     """
     Enum for the different basis sets in Qasm.
--- a/src/pyqasm/elements.py
+++ b/src/pyqasm/elements.py
@@ -120,8 +120,8 @@ class Frame:
     """
 
     port: Any
-    frequency: Any
-    phase: Any
+    frequency: float
+    phase: float
     time: Optional[Any] = 0.0
 
 
--- a/src/pyqasm/openpulse_visitor.py
+++ b/src/pyqasm/openpulse_visitor.py
@@ -1,0 +1,376 @@
+# Copyright 2025 qBraid
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pylint: disable=all
+# mypy: disable-error-code=union-attr
+"""
+Module defining OpenPulse Visitor.
+
+"""
+import logging
+from typing import Any
+
+import openpulse.ast as o_pulse_ast
+
+from pyqasm.elements import Frame
+from pyqasm.exceptions import (
+    raise_qasm3_error,
+)
+
+logger = logging.getLogger(__name__)
+logger.propagate = False
+
+
+class OpenPulseVisitor:
+    """A visitor for basic OpenPulse program elements.
+
+    This class is designed to traverse and interact with elements in an OpenPulse program.
+
+    Args:
+
+    """
+
+    def __init__(
+        self,
+        module,
+        check_only: bool = False,
+        is_def_cal: bool = False,
+    ):
+        self._module = module
+        self._check_only: bool = check_only
+        self._frames = list[Frame]
+        self._ports = list[o_pulse_ast.PortType]
+        self._waveforms = list[o_pulse_ast.WaveformType]
+        self._is_def_cal = is_def_cal
+
+    def _visit_barrier(
+        self, barrier: o_pulse_ast.QuantumBarrier
+    ) -> list[o_pulse_ast.QuantumBarrier]:
+        """Visit a barrier statement element.
+
+        Args:
+            statement (o_pulse_ast.QuantumBarrier): The barrier statement to visit.
+
+        Returns:
+            None
+        """
+        if barrier.qubits:
+            print("TODO")
+            for frm in barrier.qubits:
+                print(frm, "TODO")
+                # check frame in global scope
+
+        return []
+
+    def _visit_classical_declaration(
+        self, statement: o_pulse_ast.ClassicalDeclaration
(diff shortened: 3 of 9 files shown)


src/pyqasm/elements.py
```diff
@@ -106,6 +106,25 @@ class Variable:  # pylint: disable=too-many-instance-attributes
...
+    Attributes:
+        port (Any): The port to which the frame is attached. Immutable after initialization.
+        frequency (float): The frequency of the frame.
+        phase (float): The phase of the frame, in radians.
+        time (float): The current time of the frame, in duration units.
+                      This is manipulated implicitly and cannot be set directly.
+    """
+
+    port: Any
+    frequency: Any
```
TheGupta2012, line 123:
```suggestion
    frequency: float
```

src/pyqasm/elements.py
```diff
@@ -106,6 +106,25 @@ class Variable:  # pylint: disable=too-many-instance-attributes
...
+    phase: Any
```
TheGupta2012, line 124:
```suggestion
    phase: float
```

src/pyqasm/visitor.py
```diff
@@ -2501,6 +2505,74 @@ def _visit_include(self, include: qasm3_ast.Include) -> list[qasm3_ast.Statement
...
+            raise SyntaxError from e
+
+        ## check and update def_cal function name as gate_name
+        ## Allign virtual qubits with program qubits
+
+        result = []
+        calibration_stmts: list[openpulse_ast.Statement] = block_body.body
+        pulse_visitor = OpenPulseVisitor(module=self, is_def_cal=True)
+
+        result.extend(pulse_visitor.visit_basic_block(calibration_stmts))
```
TheGupta2012, line 2532: I like this  approach! Makes it very segregated from original visitor

src/pyqasm/openpulse_visitor.py
```diff
@@ -0,0 +1,376 @@
...
+        module,
+        check_only: bool = False,
+        is_def_cal: bool = False,
+    ):
+        self._module = module
+        self._check_only: bool = check_only
+        self._frames = list[Frame]
+        self._ports = list[o_pulse_ast.PortType]
+        self._waveforms = list[o_pulse_ast.WaveformType]
+        self._is_def_cal = is_def_cal
```
TheGupta2012, line 55: We will most likely need some kind of reference to the scope of the `QasmVisitor` object to do the semantic validations. And `self._module = module` can then be derived from the visitor object

src/pyqasm/visitor.py
```diff
@@ -98,6 +101,7 @@ def __init__(  # pylint: disable=too-many-arguments
         self._clbit_labels: dict[str, int] = {}
         self._alias_qubit_labels: dict[tuple[str, int], tuple[str, int]] = {}
         self._global_qreg_size_map: dict[str, int] = {}
+        self._calibration_qbts_map: dict[str, int] = {}  # Track qubits in defcal and cal blocks
```
TheGupta2012, line 104:
```suggestion
        self._calibration_qubit_map: dict[str, int] = {}  # Track qubits in defcal and cal blocks
```
Let's keep the full words unless we are going more than 25 chars

src/pyqasm/pulse/openpulse_visitor.py
```diff
@@ -0,0 +1,379 @@
...
+                elif isinstance(angle_arg, pulse_ast.AngleType):
+                    print("TODO")
+                else:
+                    print("ERROR")
+
+            # update frame.phase
+
+            else:
+                print("ERROR")
+        if statement.name.name == "get_phase":
```
TheGupta2012, line 242: I think we should refactor all the pulse specific keywords like `get_phase`, `shift_phase`, etc. as `Enum` or equivalent inside the  `pulse/maps.py`. It would be cleaner to identify these with specific naming conventions such as `PulseMethod.<name>`, `PulseWaveform.<name>`, etc. and refer here

src/pyqasm/pulse/openpulse_visitor.py
```diff
@@ -0,0 +1,379 @@
...
+        if visitor_function:
+            if isinstance(statement, pulse_ast.ExpressionStatement):
+                # these return a tuple of return value and list of statements
+                _, ret_stmts = visitor_function(statement)  # type: ignore[operator]
+                result.extend(ret_stmts)
+            else:
+                result.extend(visitor_function(statement))  # type: ignore[operator]
+        else:
+            raise_qasm3_error(
+                f"Unsupported statement of type {type(statement)}",
```
TheGupta2012, line 355: This will need to be clearer, that this is an unsupported statement inside the Pulse grammar. This is to ensure that users are not confused that, say, `QubitDeclaration`(I'm assuming those are not allowed in Pulse?) made inside the `cal` block is an invalid QASM statement. It should be evident from the error message that we are parsing the Pulse level description of a program and in the given context, this statement is invalid

TheGupta2012 (changes requested):
Hi @vinayswamik , I think the architecture looks good! 

Please let me know about the scoping issue , besides that we can start implementing the semantic checks

src/pyqasm/visitor.py
```diff
@@ -120,6 +124,7 @@ def __init__(  # pylint: disable=too-many-arguments
         self._in_generic_gate_op_scope: int = 0
         self._qubit_register_offsets: OrderedDict = OrderedDict()
         self._qubit_register_max_offset = 0
+        self._pulse_visitor = OpenPulseVisitor(qasm_visitor=self, check_only=check_only)
```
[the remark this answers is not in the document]
vinayswamik: Since `cal` block is consider as global scope, it's not possible to do isolated scopes. one way i suggest is same scope, as long as OpenPulseVisitor references QasmVisitor, there shouldn't be any circular dependency. 
```python
Class OpenPulseVisistor:
      def __init__(        
            self,
            qasm_visitor,
            check_only: bool = False,
             ):
                   ....
                   self._qasm_visitor = qasm_visitor
                   ....
```
This allows all scope checks and variable declarations in the pulse visitor to use the existing scope management methods from `QasmVisitor`, ensuring that both QASM and Pulse constructs share the same namespace. we can simply call methods like `_add_var_in_scope` and `_get_from_visible_scope` on `self._qasm_visitor`. 


One more idea is Create a `ScopeManager` object and pass it to both visitors. But this cost extra code in utility file and difficult to handle.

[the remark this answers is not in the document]
TheGupta2012: You raise a good point and I think going ahead with the `ScopeManager` approach would be a cleaner way to think about it because the scope is gonna be a shared entity with these visitors. 

It will surely need some analysis for separating out the scoping operations from the visitor but, once done, we can use those abstractions directly in the `Qasm` and `PulseVisitor`

.github/workflows/docs.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - name: Install dependencies
       run: |
         python3 -m pip install --upgrade pip
-        python3 -m pip install -e '.[docs]'
+        python3 -m pip install -e '.[docs,pulse]'
```
ryanhill1, line 24: pulse extra shouldn't be required to build docs. openpulse should instead just be added to [autodoc_mock_imports](https://github.com/qBraid/pyqasm/blob/da60f1c7a33ccf97cdd3f9cfe4a26e04c3aaee28/docs/conf.py#L44)

PR #233, opened 2025-07-16
Fixes #216

### LINKED ISSUE #216: [FEATURE] Implement a `.compare()` method for `QasmModule`
### Feature Description

**Signature**: `.compare(other_module)`

**Purpose**: Compares two `QasmModule` objects across multiple attributes: qubit count, depth, gate counts, external gates, user operation history, etc.

**Output**: Uses [`tabulate`](https://pypi.org/project/tabulate/) for a formatted comparison table in stdout.

#### Comparison Attributes
1. Qubit Count
2. Depth
3. Gate Counts: Per-gate-type statistics (e.g., count of CX, H, X, etc.)
4. External Gates: List of gates marked as external (e.g., from loads(external_gates=...))
5. User Operation History: A list of operations performed on the module
 
6. TBD

#### User Operation History
Helps user understand how they arrived at the current state of their module object. 

- **Initialize**: As a list property in `QasmModule` (e.g., self._user_operations = []).
- **Record**: Append operation names and params (e.g., "unroll", "validate", "segment") to the list whenever a method is called.
- **Output**: Include this list in the comparison table.

arunjmoorthy:
Fixes #216 

## Summary of changes
- in base.py added user_operations to track history of gates
- in base.py, added get_gate_counts to count all gates used
- in base.py, added compare that compares QasmModule's and shows attributes
- Added test_compare in tests/visualizations to check for proper output

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ Types of changes:
 - A new discussion template for issues in pyqasm ([#213](https://github.com/qBraid/pyqasm/pull/213))
 - A github workflow for validating `CHANGELOG` updates in a PR ([#214](https://github.com/qBraid/pyqasm/pull/214))
 - Added `unroll` command support in PYQASM CLI with options skipping files, overwriting originals files, and specifying output paths.([#224](https://github.com/qBraid/pyqasm/pull/224))
-- Added `.github/copilot-instructions.md` to the repository to document coding standards and design principles for pyqasm. This file provides detailed guidance on documentation, static typing, formatting, error handling, and adherence to the QASM specification for all code contributions. ([#234](https://github.com/qBraid/pyqasm/pull/234))s
+- Added `.github/copilot-instructions.md` to the repository to document coding standards and design principles for pyqasm. This file provides detailed guidance on documentation, static typing, formatting, error handling, and adherence to the QASM specification for all code contributions. ([#234](https://github.com/qBraid/pyqasm/pull/234))
 - Added a new `QasmModule.compare` method to compare two QASM modules, providing a detailed report of differences in gates, qubits, and measurements. This method is useful for comparing two identifying differences in QASM programs, their structure and operations. ([#233](https://github.com/qBraid/pyqasm/pull/233)) 
 
 ### Improved / Modified


TheGupta2012 (changes requested):
Thanks @arunjmoorthy for the work! The tabular format looks good, gave some comments about the requirements and code.

src/pyqasm/modules/base.py
```diff
@@ -156,6 +166,7 @@ def remove_measurements(self, in_place: bool = True) -> Optional["QasmModule"]:
         Returns:
             QasmModule: The module with the measurements removed if in_place is False
         """
+        self._user_operations.append("remove_measurements")
```
TheGupta2012, line 169: While this works, a cleaner approach would be to use decorators in Python. You can define a decorator which can be called whenever a function is executed and then append the name to the `user_operation` list , something like - 

```python
def track_user_operation(func):
    def wrapper(self, *args, **kwargs):
        self._user_operations.append(func.__name__)
        return func(self, *args, **kwargs)
    return wrapper

class QasmModule(ABC):
    # ... existing code ...

    @track_user_operation
    def validate(self):
        # ... method body ...
```

src/pyqasm/modules/base.py
```diff
@@ -16,12 +16,17 @@
...
+
+import functools
 from abc import ABC, abstractmethod
+from collections import Counter
 from copy import deepcopy
 from typing import Optional
 
 import openqasm3.ast as qasm3_ast
 from openqasm3.ast import BranchingStatement, Program, QuantumGate
+from tabulate import tabulate
```
ryanhill1, line 29: This should be imported lazily (i.e. outside top level, within the `compare` method) because it is not a core dependency, so will raise a `ModuleNotFoundError` if user does not have the `visualization` extra installed.

ryanhill1 (changes requested):

TheGupta2012:
Hi @arunjmoorthy , thanks for the updates! Besides the one change that @ryanhill1 mentioned, I think the code looks good. Can you address that and resolve merge conflicts? Should be good to merge post that

TheGupta2012:
Hi @arunjmoorthy , changes lgtm. Can you fix the format for the PR before I can merge?

TheGupta2012 (dismissed):
Lgtm

ryanhill1:
On PR squash and merge make sure to change the commit name to something more descriptive

Issue #113, opened 2024-12-14
[FEATURE] Add custom include file support 


TheGupta2012:
### Feature Description

We want users to be able to add include files other than `stdgates.inc` and `qelib1.inc` in qasm code. 

The primary motivation for this feature came out of the integration with [qBraid SDK's IONQ](https://github.com/qBraid/qBraid/pull/808#discussion_r1880489441) component. IONQ supports multiple definitions of the ms gate with 2 or 3 parameters but multiple gate definitions are not supported by openqasm. 

With this feature, we will be able to specify something like a `ionqnative.inc` file containing custom gate definitions. If the gate names are same as those identified by pyqasm, the explicitly supplied definitions will take precedence.  

#### Design Question
- We will need to see if there is an option to **switch parser context** with include files. This will enable us to just point the parser to the new include file code and continue the parsing as before. But the error reporting would be little tricky as we will have to manage something like a "file context" to know where we are at (and in which file). 

- If not above, we can also add a preprocessing step for the analyser which will make 1 big file containing all the includes and then start the analysis. 

Reference : https://openqasm.com/language/comments.html#included-files

TheGupta2012:
Closed via #236

PR #236, opened 2025-07-25
Custom include statements

LukeAndreesen:
Custom include file support ([issue](https://github.com/qBraid/pyqasm/issues/113))

A basic implementation of processing custom .inc and .qasm "include" files in QASM3 programs. This implementation simply inserts the contents of included files directly into the file where they are referenced _prior to parsing the program_. This is a preliminary, naive approach - any feedback would be appreciated.

## Summary of Changes
- Created `_process_include_statements` in [entrypoint.py](https://github.com/qBraid/pyqasm/blob/main/src/pyqasm/entrypoint.py), which simply iterates through included files, searches for the included file in user's directory, then replaces the include statement with the content of the file. 
    - We take this approach rather than attempting to switch parser context because included files are not necessarily valid QASM programs that can be parsed 
- Created additional test [circuits](https://github.com/qBraid/pyqasm/tree/luke_include/tests/qasm3/resources/qasm/custom_include) for custom included file workflow, and corresponding tests in [test_entrypoint.py](https://github.com/qBraid/pyqasm/blob/luke_include/tests/qasm3/test_entrypoint.py)
- Currently, recursive inclusion processing is not implemented, ie. if an included file includes another custom file, that is not currently processed - will implement after preliminary feedback on approach.
-


src/pyqasm/preprocess.py
```diff
@@ -0,0 +1,80 @@
...
+                            # TODO: recursive handling for nested includes
+
+                        # Replace the include line with the content
+                        program_lines[idx] = include_content
+                        processed_files.add(include_filename)
+                        include_found = True
+                        break
+                except FileNotFoundError:
+                    continue
+            if not include_found:
```
TheGupta2012, line 78: Let's move the error raise before (L57) we try to read the file in L60. This is because we already have the `include_paths` variable so just checking first will save us from the try except block.

src/pyqasm/preprocess.py
```diff
@@ -0,0 +1,80 @@
...
+    for idx, line in enumerate(program_lines):
+        line = line.strip()
+        if line.startswith("include"):
+            # Extract include filename from quotes
+            match = re.search(r'include\s+["\']([^"\']+)["\']', line)
+            if not match:
+                raise ValidationError("Invalid include statement detected in QASM file.")
+            include_filename = match.group(1)
+            # Skip stdgates.inc and already processed files
+            if include_filename == "stdgates.inc" or include_filename in processed_files:
```
TheGupta2012, line 50: I think we need to cater to `qelib1.inc` as well for qasm2 backward compatibility

src/pyqasm/preprocess.py
```diff
@@ -0,0 +1,80 @@
...
+                continue
+            # Try to find include file relative to main file first, then current directory
+            include_paths = [
```
TheGupta2012, line 53: Let's add the `.qasm` suffix restriction while detecting the valid include paths

src/pyqasm/preprocess.py
```diff
@@ -0,0 +1,80 @@
...
+
+                        # Replace the include line with the content
+                        program_lines[idx] = include_content
+                        processed_files.add(include_filename)
+                        include_found = True
+                        break
+                except FileNotFoundError:
+                    continue
+            if not include_found:
+                raise FileNotFoundError(f"Include file '{include_filename}' not found.") from None
```
TheGupta2012, line 79: Good to add more verbosity to the error messages, specifically - line number, column number and the line in which this error was found

src/pyqasm/preprocess.py
```diff
@@ -0,0 +1,80 @@
...
+    program_lines = program.splitlines()
+    processed_files = set()
+
+    for idx, line in enumerate(program_lines):
+        line = line.strip()
+        if line.startswith("include"):
+            # Extract include filename from quotes
+            match = re.search(r'include\s+["\']([^"\']+)["\']', line)
+            if not match:
+                raise ValidationError("Invalid include statement detected in QASM file.")
```
TheGupta2012, line 47: Similar comment as [below](https://github.com/qBraid/pyqasm/pull/236/files#r2234997004)

TheGupta2012 (changes requested):
Hi @LukeAndreesen , thanks for the changes! It looks good overall, some comments about tests and minor refactoring. 

Also, if you wanna give it a try, the recursive approach for multiple includes can be implemented using a stack as well.

LukeAndreesen:
Hey @TheGupta2012 , I just pushed over a fresh version of the pre-processing routine, which now uses a stack-based approach to recursively handle nested inclusions. I ended up needing to change around my last commit quite a bit to work out some of the details of this nested include handling, so apologies for a lot of fresh code. 

I tried to get things to a pretty clean and comprehensible place, but the code definitely still feels a bit confusing and perhaps not as efficient as it could be, though it does work. As such, any feedback on the code, or suggestions on a better approach to take, would be greatly appreciated!

Thanks so much,
Luke

src/pyqasm/preprocess.py
```diff
@@ -0,0 +1,202 @@
...
+def _get_custom_includes(
+    ctx: IncludeContext, current_file: str, line: str, line_idx: int
+) -> list[tuple[str, int, int]]:
+    """Extracts the custom include file name from a line.
+
+    Args:
+        line (str): The line containing the include statement.
+
+    Returns:
+        str | None: The name of the included file or None if not found.
```
ryanhill1, line 128: Update to match type annotation: `list[tuple[str, int, int]]`

src/pyqasm/preprocess.py
```diff
@@ -0,0 +1,202 @@
...
+
+def _resolve_include_path(base_file: str, file_to_include: str) -> str | None:
+    """Resolve the include path for a given file.
+
+    Args:
+        base_file (str): The base file from which the include is being made.
+        file_to_include (str): The file to include.
+
+    Returns:
+        str: The resolved include path.
```
ryanhill1, line 157: Update to match type annotation: `str | None`

src/pyqasm/preprocess.py
```diff
@@ -0,0 +1,202 @@
...
+            return path
+    return None
+
+
+def _collect_headers(ctx: IncludeContext, base_filename: str) -> None:
+    """Collects the header lines (OPENQASM and standard includes) from the base file.
+    Args:
+        base_filename (str): The base filename to read.
+    Returns:
+        list[str]: A list of header lines to include at the top of the final program.
```
ryanhill1, line 171: Update to match type annotation: `None`

ryanhill1 (changes requested):

src/pyqasm/preprocess.py
```diff
@@ -0,0 +1,203 @@
...
+            if PATTERNS["openqasm"].match(line) and line.strip() not in ctx.base_file_header:
+                # ensure openqasm line comes first
+                ctx.base_file_header.insert(0, line.strip())
+            if line.strip() not in ctx.base_file_header:
+                ctx.base_file_header.append(line.strip())
+            if PATTERNS["include_stdgates"].match(line):
+                ctx.include_stdgates = True
+            if PATTERNS["include_qelib1"].match(line):
+                ctx.include_qelib1 = True
+
```
TheGupta2012, line 190: Since `OPENQASM` header should be the first line of the program and we only need to account for the existence standard gates' files in code, we can `break` the loop once all these criteria are met i.e. - 
```python
if ctx.include_stdgates and ctx.include_qelib1 and len(ctx.ctx.base_file_header) > 0 : 
     break 
```

Just make sure to double check the length condition with some tests

TheGupta2012 (changes requested):
Nice work @LukeAndreesen ! I like the `IncludeContext` and stack based approach. Gave a few comments but looks pretty close to merging

LukeAndreesen:
Hey @TheGupta2012 , I've made some changes to ensure the insertions occur at the line of the include statement in the original script - this ended up being easier by using a recursive call rather than stack approach to avoid reading files multiple times; hopefully the logic here should be a bit clearer to follow.

tests/qasm3/test_preprocess.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2025 qBraid
```
TheGupta2012, line 1: Let's add a preprocessing test for qasm2 programs as well. It doesn't have to be an elaborate test set as we are already testing all of the processing cases in the qasm3 tests. This is just to ensure backward compatibility with qasm2 programs which might include `qelib1.inc` files.

TheGupta2012 (changes requested):
Great work @LukeAndreesen ! The recursive approach sure looks a lot cleaner. Once you add the qasm2 tests, I think the code should be good to merge.

TheGupta2012 (approved):
Lgtm

ryanhill1 (approved):

PR #257, opened 2025-08-27
Fixes: qBraid logo error in docs

vinayswamik:
## Summary of changes
- Re-uploaded qBraid logo

Closes #235

PR #258, opened 2025-08-28
Add asv benchmarking support

### LINKED ISSUE #67: [FEATURE] Create Benchmarks for unroll and validation performance
### Feature Description

Our goal for `pyqasm` is to scale its performance for large qasm programs and we want to setup a benchmarking suite to test the same. A good starting point would be to see the [benchpress repository](https://github.com/Qiskit/benchpress/tree/main/benchpress/qasm) for a vast set of qasm benchmarks. 

Ideally should aim for processing times to be <= 1s for lines <= 10k. 

### Implementation (Optional)

- A good example is [qiskit's `asv` benchmarks](https://qiskit.github.io/qiskit/) which track the package performance with each commit but might be harder to setup. Reference : [asv docs](https://asv.readthedocs.io/en/stable/using.html)
- Another lightweight tool is [`pyperf`](https://github.com/psf/pyperf) which can be used but does not have a GUI interface so it can be harder to interpret. Moreover, there is no support to see the evolution of performance as a function of the project commit history

vinayswamik:
## Summary of changes
- Added [asv](https://asv.readthedocs.io/en/stable/using.html) benchmarking support in **PyQASM**.
- Supported benchmarking functions are `import`, `load`, `draw`, `unroll`, `validate`, `dump` and ,`dumps`.
-  Also added benchmarks for `Qubit Register Consolidation` and `Openpulse` programs.

### Verification Instructions:
From the project root, you can verify the benchmarks with the following steps:

Creates a benchmarking environment under .asv/.
```bash
asv run
```
Generates `html` output files containing benchmark results in tests/benchmarking/.
```bash
asv publish
```
Launches a local server to view benchmarking results in your browser.

```bash
asv preview
```

### Notes
- Currently, the benchmark results are based on the latest commit only (HEAD). So, only a single data point is shown until historical runs are added.

### Profiling
Profiling data and visualizations for PyQASM functions
- Added SVG **flamegraphs** for `openpulse`, `qubit register consolidation`, `unroll`, and `validate` profiling.
- Included **.speedscope.json** files for detailed and interactive profiling data analysis.

Closes #67 

## Summary by CodeRabbit

- **New Features**
  - Added automated performance benchmarking across Python 3.10–3.14.
  - Added benchmarks for importing, loading, validating, unrolling, serializing, drawing, and processing OpenPulse programs.
  - Added benchmark datasets, metadata, download caching, and a neutral-atom calibration example.

- **Documentation**
  - Added an Unreleased changelog entry documenting benchmarking support.

- **Chores**
  - Added repository rules for benchmark files and downloaded benchmark assets.


ryanhill1 (changes requested):

ryanhill1 (changes requested):

ryanhill1:
Refreshed this branch to get it review-ready again (pushed three commits — main advanced mid-refresh, so there are two merge commits):

- **Merged `main`** — only `CHANGELOG.md` conflicted; kept the #258 entry under the current *Unreleased → Added* section and dropped the stale #257 line that came along from last year's changelog state.
- **`benchmark_downloader`**: removed the `CERT_NONE` SSL context + global `install_opener` (downloads now verify certificates normally — confirmed all three benchpress files fetch fine without it), and made the caching the module docstring promised real: files cache to `tests/benchmarks/qasm/<name>` (gitignored) instead of a fresh temp file per call, so repeated asv setups stop re-downloading ~600KB.
- **`asv.conf.json`**: `html_dir` → `.asv/html` so `asv publish` doesn't dirty the tree; added Python 3.14.

Verified against current `main`: all benchmark code paths run against today's API (`unroll(consolidate_qubits=True)` repeated calls are stable now — the old #296 concern no longer applies), `asv check` passes, and a real `asv run --quick` of `validate` reports ok across all three file sizes. pylint 10.00/10 on `tests/benchmarks`, black/isort clean, full suite 648 passed.

The earlier review asks (benchpress + pinned commit hash, size-parameterized benchmarks, `FileNotFoundError`/`ValueError`, lint-disable cleanup, `.asv/` in `.gitignore`) all look addressed in the September commits. One open question for reviewers: the committed profile artifacts under `tests/profiles/` (~2,000 of the PR's ~2,500 lines) are py-spy snapshots from September — worth deciding whether to keep them in-repo or regenerate/drop them before merge.

ryanhill1:
Follow-up: removed the committed py-spy artifacts under `tests/profiles/` (SVG flamegraphs + speedscope JSONs, ~2,000 lines) — they snapshot a specific commit's performance and go stale on any code change, so keeping them truthful would mean regenerating continuously. The asv benchmarks remain the reproducible record; profiles can be regenerated locally with py-spy when a specific investigation calls for one. Also trimmed the CHANGELOG entry to match ("benchmarking support", since the committed profiling data is gone). PR is now +540/−0.

PR #261, opened 2025-09-06
Fix 217: Merge function for Qasm Modules

arunjmoorthy:
## Summary of changes
Fixes issue 217 by adding a merge function, so two `QasmModule` objects can be combined into one. The merge function unrolls both modules and creates a single qubit declaration. This function appends the second `QasmModule` object with the correct index offsets and returns a Qasm3Module.

The `test_merge.py` file is added, which has a few examples of merging two modules, using the merge function.

### DIFF
--- a/src/pyqasm/modules/base.py
+++ b/src/pyqasm/modules/base.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 
 import functools
 from abc import ABC, abstractmethod
+import re
 from collections import Counter
 from copy import deepcopy
 from typing import Optional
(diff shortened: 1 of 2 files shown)


src/pyqasm/modules/base.py
```diff
@@ -761,3 +763,162 @@ def accept(self, visitor):
...
+                                start_v = int(start_s) + offset
+                                end_v = int(end_s) + offset
+                                first.name = f"__PYQASM_QUBITS__[{start_v}:{end_v}]"
+                return
+
+        # Append statements with index offset, skipping its qubit declaration and include statements
+        for stmt in right_mod.unrolled_ast.statements:
+            if isinstance(stmt, (qasm3_ast.QubitDeclaration, qasm3_ast.Include)):
+                continue
+            stmt_copy = deepcopy(stmt)
```
TheGupta2012, line 900: Don't think we'll need a copy of the statements again as we are already working on the copy of the module itself

PR #267, opened 2025-09-15
Fix 217: Merge function for Qasm Modules

arunjmoorthy:
## Summary of changes

Fixes issue 217 by adding a merge function, so two QasmModule objects can be combined into one. The merge function unrolls both modules and creates a single qubit declaration. This function appends the second QasmModule object with the correct index offsets and returns a Qasm3Module.

The test_merge.py file is added, which has a few examples of merging two modules, using the merge function.

### DIFF
--- a/src/pyqasm/modules/qasm2.py
+++ b/src/pyqasm/modules/qasm2.py
@@ -26,7 +26,7 @@ from openqasm3.printer import dumps
 from pyqasm.exceptions import ValidationError
 from pyqasm.modules.base import QasmModule
 from pyqasm.modules.qasm3 import Qasm3Module
-
+from pyqasm.modules.base import offset_statement_qubits
 
 class Qasm2Module(QasmModule):
     """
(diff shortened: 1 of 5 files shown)


src/pyqasm/modules/qasm2.py
```diff
@@ -27,6 +27,62 @@
 from pyqasm.modules.base import QasmModule
 from pyqasm.modules.qasm3 import Qasm3Module
 
+try:
+    from pyqasm.modules.base import offset_statement_qubits  # type: ignore
+except Exception:  # pylint: disable=broad-except
+
+    def offset_statement_qubits(stmt: qasm3_ast.Statement, offset: int):  # type: ignore[override]  # pylint: disable=too-many-branches
```
TheGupta2012, line 34: Why exactly are we replicating this function in both the `qasm2` and `qasm3` modules? I believe we could easily make it a method for the base module and use that as `self._offset_statement_qubits`. Or maybe I'm missing something?

src/pyqasm/modules/qasm3.py
```diff
@@ -52,3 +112,79 @@ def accept(self, visitor):
...
+        If ``other`` is QASM2, it will be converted to QASM3 before merging.
+        The merged program keeps version "3.0".
+        """
+        if not isinstance(other, QasmModule):
+            raise TypeError(f"Expected QasmModule instance, got {type(other).__name__}")
+
+        # Convert right to QASM3 if it supports conversion; otherwise copy
+        convert = getattr(other, "to_qasm3", None)
+        right_mod = (
+            convert(as_str=False) if callable(convert) else other.copy()
```
TheGupta2012, line 127:
```suggestion
        right_mod = ( other.to_qasm3() if hasattr(other, "to_qasm3") else other.copy() )
```
seems cleaner

src/pyqasm/modules/qasm2.py
arunjmoorthy: Yeah I think the way you mentioned is best. Will fix and push

PR #306, opened 2026-03-26
Fix classical register visibility inside box scope

ryanhill1:
## Summary

- Fix `box` blocks being treated as isolated scopes (like `gate`/`function`), preventing access to classical registers from the enclosing scope
- `c[1] = measure q[1];` inside a `box` block now correctly resolves the classical register `c` declared at global scope

## Details

In `scope.py`, `get_from_visible_scope()` grouped `in_box_scope()` with `in_function_scope()` / `in_gate_scope()` / `in_pulse_scope()`, which only allows access to constants and qubits from global scope. Classical registers were invisible, causing "Missing clbit register declaration" errors.

Per the OpenQASM 3.0 spec, `box` is a timing directive that executes in the enclosing scope — not an isolated scope like a gate or function definition. This PR moves `in_box_scope()` to the block scope group, which has full parent scope visibility.

Closes #302

## Summary by CodeRabbit

* **Bug Fixes**
  * Classical register declarations are now visible inside box scopes; measurements in box blocks can reference registers declared in enclosing or global scopes without validation errors.
* **Documentation**
  * Changelog updated to document the above fix.
* **Tests**
  * Added tests covering measurements to classical registers from within box scopes (nested blocks, multiple registers, multiple boxes).
* **Chores**
  * Removed a pylint suggestion-mode setting.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -104,10 +104,6 @@ recursive=no
 # source root.
 source-roots=
 
-# When enabled, pylint would attempt to guess common misconfiguration and emit
-# user-friendly hints instead of false-positive error messages.
-suggestion-mode=yes
-
 # Allow loading of arbitrary C extensions. Extensions are imported into the
 # active Python interpreter and may run arbitrary code.
 unsafe-load-any-extension=no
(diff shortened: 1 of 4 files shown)


TheGupta2012 (approved):
Changes look good

PR #320, opened 2026-06-07
Add c3x and rc3x gates, fix broken c4x, extend ctrl chain

ryanhill1:
## What

Adds the multi-controlled-X gates missing/broken in pyqasm, surfaced while transpiling qiskit's `qelib1.inc` extension gates (qBraid issue: `test_preprocess_qasm_1` is skipped because `c3x` is unsupported):

- **`c3x`** (3-controlled X) — was missing entirely. Added to `FOUR_QUBIT_OP_MAP`, decomposed into `h`/`p`/`cx` (translated from qiskit's `C3XGate.definition`).
- **`rc3x`** (relative-phase 3-controlled X, a.k.a. `rcccx`) — was missing. Added, decomposed into `h`/`t`/`tdg`/`cx`.
- **`c4x`** (4-controlled X) — was **broken**: `c4x_gate()` was declared with 4 parameters for a 5-qubit gate, so `c4x` raised `TypeError: c4x_gate() takes 4 positional arguments but 5 were given`. Reimplemented via `rc3x`/`c3sx`/`cphaseshift` (qiskit's structured decomposition).
- **`CTRL_GATE_MAP`** extended with `ccx -> c3x` and `c3x -> c4x` so the controlled-modifier chain resolves beyond 2 controls (`ctrl @ ctrl @ ctrl @ x` previously raised "Unsupported controlled QASM operation: x with 3 controls").
- **`map_qasm_ctrl_op_to_callable`** fixed: it resolved names through `CTRL_GATE_MAP` but only looked the result up in the one/two/three-qubit maps, so `c3x`/`c4x` resolved and then fell through to "Unsupported controlled QASM operation". The final lookup now also covers `FOUR_QUBIT_OP_MAP`/`FIVE_QUBIT_OP_MAP`, which is what actually makes the extended ctrl chain work.

## Why

`map_qasm_ctrl_op_to_callable` resolves controls by walking `CTRL_GATE_MAP` one step per control; the chain for X stopped at `ccx`, so 3+ controls were unsupported. And the named `c3x`/`c4x` gates from qiskit's qelib extensions weren't usable (`c3x` undefined, `c4x` crashing).

## Verification

Decompositions were checked against qiskit unitaries (`Operator`), not committed as a test dependency:

- `c3x` and `rc3x` reproduce the **exact** `C3XGate`/`RC3XGate` unitaries.
- `c4x` matches `C4XGate` **up to a global phase**, which is inherited from pyqasm's pre-existing `c3sx` implementation (same phase factor) and is physically irrelevant for the standalone gate.
- `ctrl @ ctrl @ ctrl @ x`, `ctrl @ ccx`, `ctrl(4) @ x`, and `ctrl @ c3x` all unroll to output identical to the corresponding named `c3x`/`c4x` gate.

Decompositions were generated directly from `qiskit.circuit.library.{C3XGate,RC3XGate,C4XGate}.definition`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ Types of changes:
 
 ### Fixed
 - Fixed the `c4x` (4-controlled X) gate, which previously raised `TypeError: c4x_gate() takes 4 positional arguments but 5 were given` because it was declared with four parameters for a five-qubit gate. It is now implemented via qiskit's structured `rc3x`/`c3sx`/`cphaseshift` decomposition. ([#320](https://github.com/qBraid/pyqasm/pull/320))
+- Added `inv @` (inverse modifier) support for the multi-controlled-X family: `inv @ c3x` / `inv @ c4x` resolve to the (self-inverse) gate, and `inv @ rc3x` / `inv @ rcccx` resolve to the correct relative-phase dagger. These previously raised `Unsupported / undeclared QASM operation`. ([#320](https://github.com/qBraid/pyqasm/pull/320))
+- Fixed the `ctrl @` modifier not resolving gate aliases: `ctrl @ toffoli` / `ctrl @ ccnot` (aliases of `ccx`) and `ctrl @ cnot` / `ctrl @ CX` (aliases of `cx`) now escalate controls identically to their canonical gate instead of raising `Unsupported controlled QASM operation`. ([#320](https://github.com/qBraid/pyqasm/pull/320))
 - Fixed classical register declarations not being visible inside `box` scope, causing "Missing clbit register declaration" errors for measurement statements inside box blocks. ([#306](https://github.com/qBraid/pyqasm/pull/306))
 - Fixed the backend-dependent `dt` duration unit being incorrectly relabeled as `ns` when unrolling `delay` and `box` statements without a `device_cycle_time`. Since `dt` cannot be converted to SI units without a sample rate, it is now preserved as `dt`. ([#317](https://github.com/qBraid/pyqasm/pull/317))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Types of changes:
 ## Unreleased
 
 ### Added
+- Added support for the `c3x` (3-controlled X) and `rc3x`/`rcccx` (relative-phase 3-controlled X) gates, decomposed into basis gates following qiskit's `C3XGate`/`RC3XGate` definitions. Also extended the `ctrl @` modifier chain so that 3- and 4-control stacks on `x` (e.g. `ctrl @ ctrl @ ctrl @ x`, `ctrl(4) @ x`) resolve to `c3x`/`c4x`. ([#320](https://github.com/qBraid/pyqasm/pull/320))
 
 ### Improved / Modified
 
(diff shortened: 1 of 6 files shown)


TheGupta2012 (approved):
## Summary

Nice surgical fix to a long-standing rough edge. The PR resolves the broken `c4x`, adds the missing `c3x`/`rc3x`, and closes the `ctrl @` chain past 2 controls, with numerical (not just structural) tests pinning each decomposition. The custom NumPy operator-builder in `tests/utils.py` is a clean test-dependency choice — keeps qiskit out of the test path while verifying the things structural tests can't catch.

**Verified locally:** 32 new tests + 559 existing qasm3 tests + 23 qasm2 tests all pass; `black --check` clean; the c4x global-phase claim in the PR description is accurate (phase ≈ e^{i·31π/32}) and is correctly factored out by `assert_unitary_equal`.

Approving — none of the comments below are hard blockers, but the `REV_CTRL_GATE_MAP` one is a real follow-up worth tracking.

---

### `src/pyqasm/maps/gates.py:1217-1245` — `REV_CTRL_GATE_MAP` asymmetry (Medium)

`CTRL_GATE_MAP` now contains `"ccx": "c3x"` and `"c3x": "c4x"`, but the inverse `REV_CTRL_GATE_MAP` (lines 1217-1229) was not updated with the mirror entries `"c3x": "ccx"` and `"c4x": "c3x"`. `REV_CTRL_GATE_MAP` is consumed by `src/pyqasm/printer.py:516-523` (`_mpl_draw_gate`), which peels off controls one level at a time to draw the base gate plus control dots. A user who loads a circuit containing `c3x`/`c4x` (without unrolling) and tries to visualise it via the matplotlib printer will fall through to `raise NotImplementedError(f"Unsupported gate: {name}")` at `printer.py:533`.

Strictly this is a pre-existing latent bug (the OLD `c4x_gate` was broken anyway), but by making `c3x`/`c4x` real first-class gates the PR makes the visualisation gap user-reachable. Adding `"c3x": "ccx"` and `"c4x": "c3x"` to `REV_CTRL_GATE_MAP` would keep the two maps symmetric.

**Suggestion:** Either extend `REV_CTRL_GATE_MAP` with the mirror entries (~2 lines), or document as a known follow-up. A small printer round-trip test for `c3x` would lock the symmetry in.

---

### `src/pyqasm/maps/gates.py:1192-1196` — `ST_GATE_INV_MAP` refactor scope (Low)

Collapsing the two branches (`SELF_INVERTING_ONE_QUBIT_OP_SET` and `ST_GATE_INV_MAP`) into one is semantically equivalent and more compact. However, it's unrelated to the c3x/rc3x/c4x feature, and mixing unrelated refactors into a feature PR makes the diff harder to bisect later. Non-blocking — just flagging the practice for future PRs.

---

### `src/pyqasm/maps/gates.py:838-864` — `_rc3x_dg_gate` duplication (Low)

`_rc3x_dg_gate` is the explicit inverse of `rc3x_gate` (reverse order with t↔tdg swap). The two functions are ~18 lines each and must stay in lock-step — if `rc3x_gate` is ever updated, `_rc3x_dg_gate` must follow, and the duplication makes that easy to miss.

**Suggestion:** Either derive the dagger programmatically in a follow-up (reverse the list, map each op through an inversion table), or at minimum add a comment in both functions cross-referencing the other.

---

### `src/pyqasm/maps/gates.py:1258-1287` — `map_qasm_ctrl_op_to_callable` docstring (Low)

The function now legitimately raises `ValidationError` when controls cannot be resolved (covered by `test_too_many_controls_raises`), but the docstring still only documents `Returns`. Adding a `Raises: ValidationError: when the control chain cannot resolve to a known gate.` section would close the loop.

---

### `tests/utils.py:34-130` — unitary simulator (positive)

Worth calling out: rolling a small NumPy-based unitary builder rather than pulling qiskit into the test path is exactly the right call. It keeps the test dependency surface small, gives readers a self-contained reference for what each basis gate should do, and verifies the actually-load-bearing property (matrix equality up to global phase) that structural-only tests miss. `assert_unitary_equal` normalising the global-phase ambiguity explicitly rather than papering over it is a particularly nice touch.

---

### `tests/qasm3/test_gates.py:484-500` — allowed basis set granularity (Low)

The `allowed` set in `test_multi_controlled_x_decomposition` is unioned across all four gates under test. If a future change to (say) `c3x_gate` accidentally introduces `ccx` where it shouldn't appear, the structural test will still pass. The numerical `test_multi_controlled_x_realises_mcx` catches that, but parameterising the allowed set per gate (`c3x` → `{h, p, cx}`, `rc3x`/`rcccx` → `{h, t, tdg, cx}`, `c4x` → union with `cp`) would be cheap defense-in-depth.

---

### `test_too_many_controls_raises` (positive)

Explicitly testing that `ctrl(5) @ x` raises rather than silently producing a wrong decomposition is exactly the right design boundary. Because `c4x` carries a global phase relative to ideal MCX (≈ e^{i·31π/32}), any future entry of `"c4x": "c5x"` in `CTRL_GATE_MAP` would lift that phase into a *controlled* global phase — a real relative phase. Pinning the current raise prevents a subtle future regression that wouldn't be caught by structural tests.

**Suggestion:** A one-line code comment near `CTRL_GATE_MAP` (line 1242-1244) explaining *why* the chain stops at `c4x` would help — the global-phase concern is in the PR description but not the source.

PR #330, opened 2026-07-20
Fix box duration validation to track delays per qubit timeline

### LINKED ISSUE #329: [BUG] Box duration validation sums delays across qubits instead of per qubit timeline
## Summary

Box duration validation sums the durations of **all** `delay` statements inside a `box`, regardless of which qubits they act on. Delays on disjoint qubits execute in parallel, so this over-constrains valid programs: a box whose per-qubit timelines all fit comfortably within the declared duration is rejected because the flat sum across qubits exceeds it.

The same schedule written as a single broadcast delay is accepted, so semantically identical programs get opposite verdicts depending on syntax.

## Reproduction (pyqasm 1.0.3)

```python
import pyqasm

# Two 200ns delays on DIFFERENT qubits: they run in parallel,
# wall-clock time is 200ns, which fits in box[300ns].
rejected = """
OPENQASM 3.0;
include "stdgates.inc";
qubit[2] q;
box[300ns] {
  delay[200ns] q[0];
  delay[200ns] q[1];
}
"""

# The identical physical schedule as one broadcast statement.
accepted = """
OPENQASM 3.0;
include "stdgates.inc";
qubit[2] q;
box[300ns] {
  delay[200ns] q;
}
"""

pyqasm.loads(accepted).unroll()   # OK
pyqasm.loads(rejected).unroll()   # raises ValidationError
```

Output for the first program:

```
pyqasm.exceptions.ValidationError: Total delay duration value '400.0ns' should be less than 'box[300.0ns]' duration.
```

Both programs describe the same schedule (every qubit idles 200ns inside a 300ns box), but only the broadcast form validates.

## Root cause

In `src/pyqasm/visitor.py` (current `main`):

- `_visit_delay_statement` adds every delay's duration to a single accumulator whenever inside a box scope, with no per-qubit bookkeeping (`self._total_delay_duration_in_box += duration_val`, main@`visitor.py:2876`). The increment happens once per *statement*, before qubit broadcast expansion — which is why the broadcast form counts 200ns while the per-qubit form counts 400ns.
- `_visit_box_statement` then compares this flat sum against the box duration (main@`visitor.py:2964-2974`).

## Expected behavior

Delays should be accumulated **per qubit timeline**: for each qubit, sum the durations of the delays acting on that qubit within the box, and require `max` over qubits (not the sum over all statements) to fit within the box duration. Concretely:

- `box[300ns] { delay[200ns] q[0]; delay[200ns] q[1]; }` → valid (each timeline is 200ns)
- `box[300ns] { delay[200ns] q[0]; delay[200ns] q[0]; }` → invalid (q[0]'s timeline is 400ns) — the current check gets this case right, but only incidentally
- `box[300ns] { delay[200ns] q; }` → each qubit's timeline is 200ns → valid (current behavior, correct)

A broadcast delay on a register should contribute its duration to each targeted qubit's timeline, which also fixes the per-statement/broadcast inconsistency.

## Context

Found while building demo programs for the qBraid composer: a `box` holding several qubits with per-qubit `delay`s (a hold-with-echo during classical feedback) required inflating the box duration to exceed the *sum* of parallel delays rather than the actual wall-clock time.

ryanhill1:
## Summary

Fixes #329.

Box duration validation summed the durations of **all** `delay` statements inside a `box` into a single accumulator, regardless of which qubits they target. Delays on disjoint qubits execute in parallel, so this rejected valid programs whose per-qubit timelines fit comfortably within the declared duration — while accepting the *same physical schedule* written as a single broadcast delay:

```qasm
box[300ns] { delay[200ns] q[0]; delay[200ns] q[1]; }  // rejected: "400.0ns > 300.0ns"
box[300ns] { delay[200ns] q; }                        // accepted (identical schedule)
```

## Changes

- **Per-qubit accounting** (`visitor.py`): the flat `_total_delay_duration_in_box` counter is replaced with a stack of per-box frames mapping `(register, index)` → accumulated delay on that qubit's timeline. A broadcast delay (no qubit args) now contributes to every qubit in scope, fixing the broadcast/per-qubit inconsistency. Validation compares the **busiest single qubit timeline** against the box duration.
- **Nested boxes**: when an inner box closes, it contributes its declared duration (or its per-qubit delay totals, if undeclared) to the enclosing box's timelines. Previously the global accumulator was reset to zero when an inner box closed, so the outer box silently lost all inner delay accounting.
- **Clearer error**: the validation message now names the offending qubit, e.g. `Total delay duration value '350.0ns' on qubit 'q[0]' should be less than 'box[300.0ns]' duration.`
- **Helper** (`analyzer.py`): extracted `Qasm3Analyzer.extract_qubit_key()` from `extract_duplicate_qubit()` so both share the qubit-identity logic (handles physical `$n` qubits and indexed identifiers).

## Semantics

- `box[300ns] { delay[200ns] q[0]; delay[200ns] q[1]; }` → **valid** (each timeline is 200ns)
- `box[300ns] { delay[200ns] q[0]; delay[200ns] q[0]; }` → still **invalid** (q[0]'s timeline is 400ns)
- `box[300ns] { delay[200ns] q; }` → still **valid** (each qubit's timeline is 200ns)
- `box[500ns] { box[400ns] { delay[100ns] q[0]; } delay[150ns] q[0]; }` → now **invalid** (400 + 150 = 550ns on q[0]; previously the inner box's time was dropped entirely)

## Testing

Test-driven: the four new/updated cases were written first and fail against the previous implementation (`5 failed, 15 passed` on `tests/qasm3/test_box.py` before the fix). With the fix:

- `tests/qasm3/test_box.py`: 21 passed (new: parallel-disjoint-delays regression, broadcast/per-qubit consistency, same-qubit sequential overflow, nested-box propagation; updated: the two error-message expectations now include the qubit)
- Full suite: **679 passed, 3 skipped** (pre-existing skips)
- `black`, `isort`, and `pylint` clean on changed files

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected `box` duration validation to track delay time independently for each qubit.
  * Parallel delays on separate qubits are now accepted when each timeline fits within the declared duration.
  * Nested box durations are properly included in enclosing box validation.
  * Validation errors now identify the specific qubit exceeding the allowed duration.
* **Tests**
  * Added coverage for parallel, broadcast, sequential, and nested box-delay scheduling scenarios.
* **Documentation**
  * Updated the unreleased changelog with the corrected box-duration behavior.


src/pyqasm/visitor.py
```diff
@@ -2960,19 +2977,30 @@ def _visit_box_statement(self, statement: qasm3_ast.Box) -> list[qasm3_ast.State
...
-                f"should be less than 'box[{box_duration_val}{time_unit}]' duration.",
-                error_node=statement,
-                span=statement.span,
-            )
-        self._total_delay_duration_in_box = 0
+        delay_frame = self._box_delay_frames.pop()
+        if _box_time_var and box_duration_val and delay_frame:
+            # Delays on different qubits run in parallel, so the box only
+            # needs to fit the busiest single qubit timeline.
+            max_qubit_key = max(delay_frame, key=lambda k: delay_frame[k])
```
TheGupta2012, line 2984: Why not just keep the `(name)` of the register with the largest duration in the box as the key rather than the `(name,index)` pair? Eg. - 

`box[350ns] { delay[200ns] q[0]; delay[400ns] q[1]; delay[300] q[0]; delay[600] q_new[0]; }` 

would only require `{ 'q' : 500, 'q_new' : 600 }` to determine whether the operations fit in the `box` duration. We can probably just make a simple `dict` rather than list of `dict`. 

Do you see a use case for the per-qubit durations in any later processing stages?

TheGupta2012 (commented):
Thanks for fixing this @ryanhill1 ! Gave one comment regarding the data structure being used, but the idea is correct

src/pyqasm/visitor.py
ryanhill1: The per-qubit keys are load-bearing — the `500` in your example is `q[0]`'s sequential total (200 + 300), i.e. the max *over per-qubit sums*. A register-keyed dict has nowhere to hold `q[0] = 200` while `q[1] = 400` arrives, so at the third delay there's no way to know which running total the `300` belongs to. Both collapse strategies lose it:

- sum into `'q'` → 200+400+300 = 900 (that's the bug this PR fixes)
- max into `'q'` → max(200, 400, 300) = 400, never 500

The max variant also regresses a test in this PR:

```qasm
box[300ns] { delay[200ns] q[0]; delay[200ns] q[1]; delay[150ns] q[0]; }
```

`q[0]`'s timeline is 350ns and must be rejected; register-max gives 200ns and silently accepts it — a false accept, worse than the false reject we started with.

The register-level max is fine as an *output*, just not as an *accumulator* — and that's what the code already does: it accumulates per qubit, then takes the max at box close for the comparison and the error message.

On `list[dict]`: that's a stack, not parallel data — one frame per *open* box, so a nested box doesn't clobber its parent's timelines. Depth is bounded by box nesting. Folding depth into a single dict's key would be strictly worse.

On later stages: I'd argue it's required for correctness here regardless of future use. That said, per-qubit timelines are the natural input to anything scheduling-related later (stretch resolution, inferring an undeclared box duration), so I don't expect this to be throwaway state.

TheGupta2012: Okay, I was considering that the value is the _max of the per-qubit sum of a register_ but I get that it might be a little complex to handle nested boxes that way

TheGupta2012 (approved):

PR #332, opened 2026-07-22
fix: remove_idle_qubits KeyError on shared operand nodes

### LINKED ISSUE #331: remove_idle_qubits() KeyError: unroll() emits aliased operand nodes that _remap_qubits mutates repeatedly
## Summary

`QasmModule.remove_idle_qubits()` raises `KeyError` on any program whose unrolled AST contains **aliased operand nodes** — the same `IndexedIdentifier` object reused across multiple statements — whenever removing an idle qubit shifts the indices of the qubits those statements use. The minimal trigger is a single `crz` with an idle lower-indexed qubit.

Found while testing the statevector simulator work in #316, but the bug is in `modules/base.py` and pre-dates that branch (reproduces on `main` / released 1.0.3).

## Reproduction

```python
from pyqasm import loads

m = loads('''OPENQASM 3;
include "stdgates.inc";
qubit[3] q;
crz(0.5) q[1], q[2];
''')
m.unroll()
m.remove_idle_qubits()   # KeyError: 0
```

```
File "src/pyqasm/modules/base.py", line 387, in _remap_qubits
    bit.indices[0][0].value = idx_map[old_idx]
KeyError: 0
```

## Root cause

`unroll()`'s decomposition of `crz` emits gate statements whose qubit operands are **the same `IndexedIdentifier` object**, not copies. For the program above, one `q[2]` node is shared by 11 statements and the `cx` operand pair is shared by both `cx` gates:

```python
from collections import Counter
gates = [s for s in m._unrolled_ast.statements if type(s).__name__ == "QuantumGate"]
Counter(id(q) for s in gates for q in s.qubits).most_common(2)
# [(<id>, 12), (<id>, 2)]   <- one operand object reused 12 times
```

`_remap_qubits` (base.py:380-387) then walks every statement and rewrites `bit.indices[0][0].value` **in place**. A node shared by k statements is remapped k times: with `idx_map = {1: 0, 2: 1}` the shared `q[2]` operand goes `2 -> 1` on the first visit, `1 -> 0` on the second, and the third visit looks up `idx_map[0]` -> `KeyError: 0`.

When the remap happens to be the identity (no lower-indexed idle qubit), the repeated remapping is silently harmless, which is why the bug stayed hidden.

## Notes

- Any decomposition that reuses operand nodes triggers this; `crz` is just the minimal known case (`cx`, `swap`, `ccx` unroll with fresh nodes and are unaffected).
- Two possible fixes, not mutually exclusive:
  1. `unroll()` should emit deep-copied operand nodes (aliasing in an AST that later gets mutated in place is a footgun beyond this one crash);
  2. `_remap_qubits` should be idempotent per node, e.g. track visited node ids or rebuild operands instead of mutating shared ones.
- Also affects the `Simulator.run(str)` convenience path in #316, which calls `remove_idle_qubits()` internally.

SAY-5:
## Summary of changes

Closes #331.

`unroll()` can emit gate statements that share the same `IndexedIdentifier` operand node (the `crz` decomposition does this), and `_remap_qubits` rewrites `bit.indices[0][0].value` in place per statement, so a node shared by k statements gets remapped k times. With a lower-indexed idle qubit the repeated lookups walk off the index map and raise `KeyError`, e.g. `crz(0.5) q[1], q[2];` on a 3-qubit register.

This tracks visited node ids so each operand node is remapped exactly once, and adds a regression test with the minimal repro from the issue. The remap now also produces correct indices for shared nodes rather than silently double-remapping them when the shift happens to stay inside the map.


ryanhill1 (dismissed):
Thanks @SAY-5 — this is a clean, minimal fix. Approving.

Verification done during review:

- Reproduced the `KeyError: 0` from #331 on `main` and confirmed this branch fixes it.
- Checked the remapped output is semantically correct: the resulting 2-qubit circuit is unitarily equivalent to `crz(0.5)` via qiskit's `Operator.equiv`.
- Adversarial probes all pass: calling `remove_idle_qubits()` twice, `in_place=False` (deepcopy memoization preserves the aliasing, so the guard still works on copies), multiple registers each with idle qubits and aliased operands, a custom gate spanning two registers, and measurements mixed in. Also confirmed `unroll()` never shares an index-literal node across different registers, so scoping `visited_node_ids` per `_remap_qubits` call is sound.
- Full test suite passes locally (the handful of failures in my env are a missing optional `tabulate` dep and reproduce on `main`).

One optional, non-blocking suggestion: the sibling tests in `test_transformations.py` compare the full unrolled program via `check_unrolled_qasm`. Doing the same here would also lock in the rotation-gate indices — i.e. the "silent double-remap when the shift stays inside the map" case from the PR description, which the current `in`/`not in` assertions don't pin down. Fine to land as-is or as a follow-up.

Heads-up: while reviewing I found the same root cause bites `reverse_qubit_order()` — same repro, `KeyError: -1` from the negative-marker pass reading an already-marked shared node. Filed as #333 (with the same two fix options as #331; the deep-copy-at-unroll route would retire the whole bug class). If you're interested in picking that one up too, it's yours — the mechanics are exactly what you just fixed here. Also filed the unrelated pre-existing #334 spotted nearby.

ryanhill1 (approved):
Re-approving after the stale-review dismissal — the substantive change is unchanged from my original review; the two new commits are the main merge (CHANGELOG conflict with #330) and the one-line pylint R0914 fix.

PR #335, opened 2026-07-24
fix: emit fresh operand nodes from gate decompositions

### LINKED ISSUE #333: reverse_qubit_order() KeyError on aliased operand nodes (same root cause as #331)
## Summary

`QasmModule.reverse_qubit_order()` raises `KeyError` on any program whose unrolled AST contains aliased operand nodes — the same root cause as #331, surfacing one method below the fix in #332.

Found during adversarial review of #332: the visited-set guard added there fixes `_remap_qubits`, but `reverse_qubit_order` (`src/pyqasm/modules/base.py:533-555`) walks the same statements and mutates `bit.indices[0][0].value` in place with a negative-marker trick. On a node shared by k statements, the second visit reads the already-negated value and looks up `new_qubit_mappings[reg][-2]` → `KeyError: -1`.

## Reproduction

```python
from pyqasm import loads

m = loads('''OPENQASM 3;
include "stdgates.inc";
qubit[3] q;
crz(0.5) q[1], q[2];
''')
m.unroll()
m.reverse_qubit_order()   # KeyError: -1
```

Reproduces on `main` (pre-dates #332; that PR did not touch this method).

## Fix options

Same two options as #331, not mutually exclusive:

1. Root cause: have `unroll()` / the decomposer emit fresh (deep-copied) operand nodes so no AST node is shared across statements. This retires the whole class of bug — every in-place index mutation in `modules/base.py` is currently exposed to aliasing.
2. Point fix: apply the same visited-node-id guard from #332 to both passes in `reverse_qubit_order`. Note the marker pass and the unmark pass each need their own guard (or the method should be rewritten to build the mapping without the negative-marker trick, which only exists to survive double visits it doesn't actually survive).

Option 1 is the better long-term fix; if we take it, the guard added in #332 becomes harmless belt-and-braces.

ryanhill1:
## Summary of changes

Closes #333. Root-cause fix for the operand-node aliasing behind #331 as well.

Gate decomposition functions in `maps/gates.py` pass the same qubit operand nodes into every statement they emit (e.g. `crz_gate` hands `qubit1` to two `u3_gate` and two `cx` expansions), so the unrolled AST contained `IndexedIdentifier` objects shared across many statements. Any transformation that rewrites qubit indices in place then mutated a shared node once per referencing statement:

- `remove_idle_qubits()` walked off its index map → `KeyError` (#331, point-fixed by the visited-node guard in #332);
- `reverse_qubit_order()`'s negative-marker pass read an already-marked shared node → `KeyError: -1` (#333, no guard);
- `Decomposer._get_decomposed_gates` had the same pattern, so `rebase()` output was aliased too.

This PR fixes the class of bug at the source: the five statement constructors in `maps/gates.py` (`one_qubit_gate_op`, `one_qubit_rotation_op`, `two_qubit_gate_op`, `ccx_gate_op`, `global_phase_gate`) now copy their qubit operands via a shared `fresh_qubits` helper, and `Decomposer` does the same for the operands of each rule-emitted gate. Every decomposition composes these constructors, so no statement emitted by a decomposition can share a node with another statement or with the source operation.

Scope note: the `negctrl` modifier expansion in `visitor.py` is an independent aliasing source that this PR does not touch — it re-inserts the *same* `QuantumGate` object before and after the controlled gate. That is pre-existing on `main` and tracked separately in #350. The `visited_node_ids` guard added in #332 is what keeps `remove_idle_qubits()` working on those circuits, so it remains load-bearing rather than redundant.

`fresh_qubits` rebuilds the (small, fully known) operand node tree by hand rather than calling `deepcopy` — a shallow `copy` is not sufficient because transforms mutate the nested `IntegerLiteral` in place. Microbenchmark over 200k iterations on `q[3]`: `deepcopy` 1.550 s, shallow `copy` 0.182 s (and incorrect), structural rebuild 0.110 s. `deepcopy` is kept only as a fallback for operand shapes that are not a plain `reg[int]`.

Verified beyond the test suite: the reversed/remapped `crz` circuits are unitarily equivalent to the expected reference circuits via qiskit `Operator.equiv`.

## Tests

- `test_reverse_qubit_order_gate_decomposition`: the #333 repro, checked against the full expected unrolled program.
- `test_unroll_emits_fresh_operand_nodes`: invariant test asserting no two quantum statements share an operand or index node after `unroll()`, parametrized over `crz`, `crx`, `c4x`, `ecr`, and `inv @ crz`.
- `test_rebase_emits_fresh_operand_nodes`: same invariant for `rebase()`, parametrized over `swap` and `cz` (which route through `Decomposer._get_decomposed_gates`) and `crz` (which does not).
- `test_fresh_qubits_shares_no_nodes`: covers the plain `reg[int]`, physical-qubit, and `deepcopy`-fallback paths of the helper.


src/pyqasm/maps/gates.py
```diff
@@ -30,6 +31,17 @@
...
 
+def _fresh_qubits(*qubits: IndexedIdentifier) -> list[IndexedIdentifier | Identifier]:
+    """Deep-copy qubit operands so every emitted statement owns its nodes.
+
+    Decomposition functions pass the same operand nodes to each statement they
+    emit; without copies, transforms that later rewrite qubit indices in place
+    (e.g. ``remove_idle_qubits``, ``reverse_qubit_order``) would mutate a
+    shared node once per referencing statement.
+    """
+    return [deepcopy(qubit) for qubit in qubits]
```
TheGupta2012: Does an `IndexedIdentifier` have nested statements as well? If not, maybe a shallow `copy` can work here too. The performance of `deepcopy` is known to be a little slow, and since this `_fresh_qubits` operation will be used quite frequently, it is best to figure out the optimized approach for copying

src/pyqasm/decomposer.py
```diff
@@ -133,7 +135,9 @@ def _get_decomposed_gates(cls, decomposition_rules, statement, gate):
                 modifiers=[],
                 name=qasm3_ast.Identifier(name=rule["gate"]),
                 arguments=arguments,
-                qubits=qubits,
+                # copy so the emitted gates do not share operand nodes with each
+                # other or with the source statement (see #333)
+                qubits=[deepcopy(qubit) for qubit in qubits],
```
TheGupta2012: Why not use `_fresh_qubits` here too?

TheGupta2012 (commented):
Thanks for flagging this @ryanhill1 , can you see if `copy` works here as well? If yes, best to use that instead of `deepcopy`.

src/pyqasm/maps/gates.py
ryanhill1: Good catch on the cost — but a shallow `copy` unfortunately isn't enough here.

`IndexedIdentifier` has no nested *statements*, but it does have a nested node tree: `IndexedIdentifier.name` is an `Identifier`, and `.indices` is a `list[list[Expression]]` holding the `IntegerLiteral`. The transforms mutate the innermost node in place:

```python
bit.indices[0][0].value = idx_map[old_idx]   # _remap_qubits
bit.indices[0][0].value = -1 * new_reg_idx - 1  # reverse_qubit_order
```

so `copy.copy` leaves the very object that gets mutated shared:

```python
s = copy(q)
s.indices[0][0].value = 99
q.indices[0][0].value  # 99  <- still aliased
```

What does work, and is faster than both, is rebuilding the node tree by hand. At this point in the pipeline every operand is either a plain `reg[int]` or a bare physical-qubit `Identifier`, so the copy is three tiny constructor calls with no reflection:

```python
def _copy_qubit(qubit):
    if isinstance(qubit, Identifier):
        fresh = Identifier(name=qubit.name)
    else:
        indices = qubit.indices
        if not (len(indices) == 1 and isinstance(indices[0], list)
                and len(indices[0]) == 1 and isinstance(indices[0][0], IntegerLiteral)):
            return deepcopy(qubit)   # fallback for any other operand shape
        fresh = IndexedIdentifier(
            name=Identifier(name=qubit.name.name),
            indices=[[IntegerLiteral(value=indices[0][0].value)]],
        )
    fresh.span = qubit.span
    return fresh
```

Microbenchmark, 200k iterations on `q[3]`:

| approach | time |
| --- | --- |
| `deepcopy` | 1.550 s |
| `copy` (insufficient) | 0.182 s |
| structural rebuild | 0.110 s |

~14x faster than `deepcopy`, and still faster than the shallow copy it can't use. `deepcopy` is kept only as a fallback for operand shapes that aren't a plain `reg[int]` (e.g. a `DiscreteSet` or multi-dimensional index), so correctness doesn't depend on that assumption holding forever.

`span` is carried over explicitly so error reporting on emitted statements is unchanged.

Added `test_fresh_qubits_shares_no_nodes` covering the plain, physical-qubit, and fallback paths.

src/pyqasm/decomposer.py
ryanhill1: Done — good call. Renamed the helper to `fresh_qubits` (dropping the leading underscore now that it crosses module boundaries) and `_get_decomposed_gates` now calls it:

```python
qubits=fresh_qubits(*qubits),
```

`decomposer.py` no longer imports `deepcopy` at all.

TheGupta2012 (commented):
**Approved with comments** — the right fix at the right level: #331 and #333 were two symptoms of one defect, and this fixes the source rather than guarding each consumer. One test gap worth closing before merge; the other notes are follow-ups.

### Findings

- [ ] M1 — `decomposer.py` hunk is unpinned; reverting it leaves the suite green
- [ ] M2 — `negctrl` path still aliases statement objects; PR body over-promises (follow-up)
- [ ] L1 — `#332` guard comment is now untrue and invites deletion of load-bearing code
- [ ] L2 — Test helper assumes an indexed operand shape

### How it was tested

- Repro: `crz(0.5) q[1],q[2]` + `unroll()` + `reverse_qubit_order()` raises `KeyError: -1` on `main`, passes here.
- Aliasing invariant: 0 shared operand/index nodes after `unroll()` across `crz`, `crx`, `ecr`, `inv @ crz`, `ccx`, `c4x`.
- Reverting `qubits=fresh_qubits(*qubits)` in `decomposer.py` alone, leaving `gates.py` intact, still leaves the full suite green (see M1).
- Removing the `#332` guard from the branch: `crz` fine, `negctrl @ x` raises `KeyError: 0`, suite still green (see L1).
- Instrumenting `Decomposer._get_decomposed_gates` to confirm which gates reach it: `crz` → no calls, `swap` → one call.
- Microbenchmark reproduced (200k iterations on `q[3]`): `deepcopy` 1.499 s, shallow `copy` 0.168 s, `_copy_qubit` 0.112 s. Mutating a shallow copy's `indices[0][0].value` writes through to the original; `_copy_qubit`'s result does not.
- Perf end to end: 200 `crz` gates → 2402 statements in 0.046 s; the `deepcopy` fallback is not on the hot path.
- Suite on branch: **647 passed, 4 skipped**. Branch current with `main` (0 behind), `MERGEABLE`.
- The parametrized test suggested in M1 was applied and run both ways: 3 passed with the fix; reverting `decomposer.py` fails the `swap` and `cz` cases and passes `crz`.

### Next steps

Add the `swap`/`cz` case from M1 to the rebase test — that is the one ask before merge; it is attached as an applicable suggestion. M2 and L1 want follow-up issues (and L1 a one-line comment edit), L2 is optional. Independent of the rest of the open batch, so it can land at any point in the sequence.

src/pyqasm/maps/gates.py
```diff
@@ -19,16 +19,66 @@
...
+            return deepcopy(qubit)
+        fresh = IndexedIdentifier(
+            name=Identifier(name=qubit.name.name),
+            indices=[[IntegerLiteral(value=indices[0][0].value)]],
+        )
+    fresh.span = qubit.span
+    return fresh
+
+
+def fresh_qubits(*qubits: QubitOperand) -> list[QubitOperand]:
```
TheGupta2012, line 72: **[M2] `negctrl` path still aliases statement objects** — Implementation · Medium *(pre-existing; follow-up issue, not a blocker)*

**Rationale:** this helper delivers what it promises for decomposition-emitted statements. But the PR description says "no emitted statement can share a node with another statement or with the source operation", and the unroller has an independent aliasing source this PR does not touch, in `visitor.py`:

```python
negs = [qasm3_ast.QuantumGate([], qasm3_ast.Identifier("x"), [], [ctrl]) for ctrl in negctrls]
result = negs + result + negs
```

That places the *same* `QuantumGate` object at two positions in the statement list, not two equal copies. Verified on this branch:

```
negctrl @ x q[0], q[1];  ->  x q[0];  cx q[0], q[1];  x q[0];
statement object ids: [4337616448, 4336546384, 4337616448]
```

So `reverse_qubit_order()` still walks the same operand node twice and raises `KeyError: -2`. Identical behaviour on `main` (`unroll()` + `reverse_qubit_order()` on `negctrl @ x` fails on both), so this is **not a regression from this PR** — it is outside the stated scope, and no test covers it on either side.

**Change requested:** nothing in this PR. Open a follow-up for the `negctrl` path (presumably a second `negs` list rather than reusing the first, plus `fresh_qubits` on `ctrl`), and consider narrowing the sentence in the PR body to "decomposition no longer emits shared nodes" so the changelog does not over-promise.

src/pyqasm/decomposer.py
```diff
@@ -133,7 +133,9 @@ def _get_decomposed_gates(cls, decomposition_rules, statement, gate):
                 modifiers=[],
                 name=qasm3_ast.Identifier(name=rule["gate"]),
                 arguments=arguments,
-                qubits=qubits,
+                # copy so the emitted gates do not share operand nodes with each
+                # other or with the source statement (see #333)
+                qubits=fresh_qubits(*qubits),
```
TheGupta2012, line 138: **[L1] The `#332` guard comment is now untrue — and the guard is still load-bearing** — Maintenance · Low

*Concerns the `visited_node_ids` guard in `_remap_qubits`, `src/pyqasm/modules/base.py` — not changed by this PR, hence the anchor here.*

**Rationale:** the PR body describes that guard as "useful belt-and-braces" now that nodes are fresh. That undersells it, and the understatement is the risk — it invites a future reader to delete it as dead code. Removing the guard from this branch and re-running gives:

```
crz         : remove_idle_qubits OK          <- this PR did make the guard redundant here
negctrl @ x : remove_idle_qubits KeyError: 0 <- still the only thing holding this up
647 passed, 4 skipped                        <- no test would have caught it
```

So the guard is precisely what keeps `remove_idle_qubits()` working on the `negctrl` circuits from M2, and nothing in the suite protects it.

**Change requested:** update the comment above `visited_node_ids` to name the surviving case (statements re-inserted by the `negctrl` expansion), rather than "gate decompositions can reuse the same operand node" — which this PR has now made untrue. One line, but it is the difference between the next reader keeping it and removing it.

tests/qasm3/test_transformations.py
```diff
@@ -156,6 +163,123 @@ def test_reverse_qubit_order_qasm3():
...
+    cx q[1], q[0];
+    """
+
+    module = loads(qasm3_str)
+    module.unroll()
+    module.reverse_qubit_order()
+    check_unrolled_qasm(dumps(module), expected_qasm3_str)
+
+
+def _assert_no_shared_operand_nodes(module):
```
TheGupta2012, line 200: **[L2] Helper assumes an indexed operand shape** — Maintenance · Low

**Rationale:** `_assert_no_shared_operand_nodes` does `bit.indices[0][0]` unconditionally, so it would raise on a bare-`Identifier` physical qubit or a `DiscreteSet` index rather than reporting a clean assertion failure. Not hit by the current parameters, and `test_fresh_qubits_shares_no_nodes` covers those shapes directly.

**Change requested:** none required. Worth a guard only if this helper gets reused for physical-qubit circuits later — noted so the limitation is known rather than rediscovered.

tests/qasm3/test_transformations.py
```diff
@@ -156,6 +163,123 @@ def test_reverse_qubit_order_qasm3():
...
+def test_rebase_emits_fresh_operand_nodes():
+    """Test that rebase() never emits statements sharing operand nodes (issue #333)"""
+    qasm3_str = """
+    OPENQASM 3.0;
+    include "stdgates.inc";
+    qubit[3] q;
+    crz(0.5) q[1], q[2];
+    """
+    module = loads(qasm3_str).rebase(BasisSet.ROTATIONAL_CX)
+    _assert_no_shared_operand_nodes(module)
```
TheGupta2012: **[M1] This test does not reach the `decomposer.py` hunk** — Implementation · Medium

**Rationale:** `test_rebase_emits_fresh_operand_nodes` uses `crz`, which is not in `DECOMPOSITION_RULES` (`ROTATIONAL_CX` covers `x, y, z, h, s, t, sx, sdg, tdg, cz, swap`). Instrumenting `Decomposer._get_decomposed_gates` confirms the test never reaches it:

```
crz  (the test in this PR) -> _get_decomposed_gates calls: []
swap (suggested)           -> _get_decomposed_gates calls: ['swap']
```

`crz` is handled entirely by the `maps/gates.py` constructors, so this rebase test is really a second unroll test. Consequence: reverting `qubits=fresh_qubits(*qubits)` back to `qubits=qubits` in `decomposer.py`, leaving `gates.py` intact, still leaves the **entire suite passing — 647 passed, including this test**. Half the diff is currently unpinned.

The hunk is genuinely fixing something, to be clear. On `main`, `swap q[0], q[2]` rebased to `CLIFFORD_T` emits three `cx` statements sharing both index nodes:

```
cx [('q', 0, 4345354976), ('q', 2, 4345355792)]
cx [('q', 2, 4345355792), ('q', 0, 4345354976)]
cx [('q', 0, 4345354976), ('q', 2, 4345355792)]
```

4 duplicate operand nodes and 4 duplicate index nodes on `main`, 0 on this branch. `_get_qubits_for_gate` hands back references straight into `statement.qubits`, so every rule-emitted gate aliases the source statement. The fix is correct; it just isn't tested.

**Change requested:** parametrize over a gate that actually reaches `_get_decomposed_gates`. The suggestion below was applied and run both ways — 3 passed with the fix in place, and reverting the `decomposer.py` line fails the `swap` and `cz` cases while `crz` still passes.

```suggestion
@pytest.mark.parametrize(
    "operation", ["crz(0.5) q[1], q[2];", "swap q[0], q[2];", "cz q[1], q[2];"]
)
def test_rebase_emits_fresh_operand_nodes(operation):
    """Test that rebase() never emits statements sharing operand nodes (issue #333)

    ``swap`` and ``cz`` are in DECOMPOSITION_RULES so they exercise
    Decomposer._get_decomposed_gates; ``crz`` is handled by maps/gates.py instead.
    """
    qasm3_str = f"""
    OPENQASM 3.0;
    include "stdgates.inc";
    qubit[3] q;
    {operation}
    """
    module = loads(qasm3_str).rebase(BasisSet.ROTATIONAL_CX)
    _assert_no_shared_operand_nodes(module)
```

> Note for whoever applies it: don't reach for `rebase(...)` followed by `reverse_qubit_order()` as the regression test — the rebase is silently discarded before the remap runs, so that test would pass for the wrong reason.

ryanhill1: Resolved in 6620dbe — suggestion applied as written.

Confirmed your finding before and after: with `qubits=fresh_qubits(*qubits)` reverted in `decomposer.py`, the `swap` and `cz` cases fail and `crz` still passes; with the fix in place all three pass. The hunk is pinned now.

src/pyqasm/maps/gates.py
ryanhill1: Acknowledged — follow-up filed as #350, PR body narrowed.

Reproduced on this branch: statement object ids for `negctrl @ x q[0], q[1];` come back `[4362757584, 4362758224, 4362757584]`, so positions 0 and 2 are the same object. Left out of scope here since it's pre-existing on `main` and lives in the modifier expansion, not decomposition.

The PR body now says "no statement emitted by a decomposition can share a node", with an explicit scope note pointing at #350.

src/pyqasm/decomposer.py
ryanhill1: Resolved in 6620dbe — comment rewritten to name the surviving case.

Verified the guard is still load-bearing: `remove_idle_qubits()` on `negctrl @ x` passes only because of it. The comment now reads:

```python
# the negctrl expansion in the visitor re-inserts the same QuantumGate object before
# and after the controlled gate, so a single index node can be reached more than once;
# track visited nodes so it is remapped exactly once
```

Cross-referenced from #350 so the two stay linked.

tests/qasm3/test_transformations.py
ryanhill1: Resolved in 6620dbe — guard added, since it was cheap.

`_assert_no_shared_operand_nodes` still checks operand-node identity for every bit, but skips the index check for bare `Identifier` operands and non-plain index shapes rather than raising on `bit.indices[0][0]`.

TheGupta2012 (dismissed):
**Approved.**

Re-verified at `6620dbe`: the parametrized rebase test is load-bearing — reverting `qubits=fresh_qubits(*qubits)` in `decomposer.py` fails the `swap` and `cz` cases while `crz` still passes, so the hunk is pinned. The `#333` repro passes, the aliasing invariant holds across all six gate shapes, and the remap-guard comment now names the surviving `negctrl` case. Suite 649 passed, 4 skipped; CI 26/26.

M2 tracked in #350, with the PR body narrowed to match.

TheGupta2012 (approved):
Re-approving after merging `origin/main` (now carrying #345 and #344) into the branch. Conflict was `CHANGELOG.md` only — both entries kept, verified nothing dropped from either side. `modules/base.py` auto-merged cleanly: #335's remap-guard comment sits alongside #345's `iter_quantum_statements` walker. Suite on the merged tree: 660 passed, 4 skipped.

PR #336, opened 2026-07-24
fix: remove_idle_qubits(in_place=False) updating qubit count on the wrong module

### LINKED ISSUE #334: remove_idle_qubits(in_place=False) decrements num_qubits on the original module instead of the copy
## Summary

`remove_idle_qubits(in_place=False)` updates the qubit count on the wrong module: the **original** module's `num_qubits` is decremented while the returned copy keeps the stale pre-removal count. The copy's AST is remapped correctly (declaration size, operand indices, `dumps()` output are all right) — only the counter is wrong, and it's wrong on both objects.

## Reproduction

```python
from pyqasm import loads

m = loads('''OPENQASM 3;
include "stdgates.inc";
qubit[3] q;
h q[1];
cx q[1], q[2];
''')
m.unroll()
m2 = m.remove_idle_qubits(in_place=False)
print(m2.num_qubits)  # 3  (should be 2)
print(m.num_qubits)   # 2  (should still be 3 — original was not supposed to change)
```

## Root cause

`src/pyqasm/modules/base.py:494` (inside `remove_idle_qubits`):

```python
self._num_qubits -= len(idle_indices)
```

Every other mutation in the method goes through `qasm_module` (which is `self.copy()` when `in_place=False`); this one line uses `self`. Line 474 (`size = self._qubit_registers[reg_name]`) also reads from `self` — harmless today since the copy's registers are identical at that point, but it should be `qasm_module` for consistency.

Found during adversarial review of #332; reproduces on `main` and is unrelated to that PR's change.

ryanhill1:
## Summary of changes

Closes #334.

`remove_idle_qubits` routes every mutation through `qasm_module` — which is `self` or a deepcopy depending on `in_place` — except the qubit-count update at the end of the register loop, which used `self`. With `in_place=False` the original module's `num_qubits` was decremented and the returned copy kept the stale pre-removal count:

```python
m2 = m.remove_idle_qubits(in_place=False)
m2.num_qubits  # 3 (stale — should be 2)
m.num_qubits   # 2 (original mutated — should still be 3)
```

The copy's AST was already correct (declaration size, operand indices, `dumps()` output); only the counters were swapped. This also switches the `size` lookup two lines above from `self` to `qasm_module` for consistency — harmless today since the registers are identical at that point, but it keeps the method reading from one object.

Regression test asserts both sides: the returned copy reports the reduced count and dumps the collapsed register, and the original module is left untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


TheGupta2012 (dismissed):

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ Types of changes:
 ### Removed
 
 ### Fixed
+- Fixed `remove_idle_qubits(in_place=False)` updating the qubit count on the wrong module: the original module's `num_qubits` was decremented while the returned copy kept the stale pre-removal count. The copy's AST was already correct; only the counters were swapped. ([#334](https://github.com/qBraid/pyqasm/issues/334))
```
TheGupta2012:
```suggestion
- Fixed `remove_idle_qubits(in_place=False)` updating the qubit count on the wrong module: the original module's `num_qubits` was decremented while the returned copy kept the stale pre-removal count. The copy's AST was already correct; only the counters were swapped. ([#336](https://github.com/qBraid/pyqasm/pull/336))
```

TheGupta2012 (approved):

PR #341, opened 2026-08-04
feat: support #pragma statements and preserve braket verbatim boxes

### LINKED ISSUE #340: pyqasm rejects #pragma braket verbatim (Pragma statements), blocking verbatim compilation through qBraid
## Summary

`pyqasm` rejects any `#pragma` statement with `ValidationError: Unsupported statement of type <class 'openqasm3.ast.Pragma'>`. This blocks **verbatim / fixed-physical-qubit execution** on Braket QPUs (IQM, Rigetti) through the qBraid stack: submitting a verbatim circuit serializes to OpenQASM 3 containing `#pragma braket verbatim`, and pyqasm rejects it *before* the program can be submitted.

## Repro (pyqasm 1.0.4)

```python
import pyqasm

qasm = '''OPENQASM 3.0;
bit[2] c;
#pragma braket verbatim
box {
prx(1.5707963267948966, 4.71238898038469) $1;
cz $2, $1;
}
c[0] = measure $2;
'''
m = pyqasm.loads(qasm)
m.unroll()
# pyqasm.exceptions.ValidationError: Unsupported statement of type <class 'openqasm3.ast.Pragma'>  (line 3)
```

## Root cause

`QasmVisitor.visit_statement` (`src/pyqasm/visitor.py:~3288`) dispatches on `type(statement)` via `self._visit_map`. `ast.Pragma` has no entry, so it falls through to the `else` and raises `Unsupported statement of type ...`. The README already lists **Pragma** as *Planned*.

## Impact

Verbatim submission through `qbraid.runtime.QbraidProvider.run` fails **client-side** for *both* input forms (verified 2026-08-04):
- a raw OpenQASM 3 string containing `#pragma braket verbatim`, and
- a native `braket.circuits.Circuit` built with `add_verbatim_box(...)` — the SDK serializes it to OpenQASM 3 with the same pragma.

Verbatim is the only Braket mechanism for fixed physical qubits (the old `disable_qubit_rewiring` kwarg was removed from `amazon-braket-sdk` by 1.120). Note the downstream OpenQASM→Braket conversion (`Circuit.from_ir`) already **preserves** verbatim boxes — pyqasm is the sole blocker.

## Proposed bypass

Add a pass-through path for `ast.Pragma`:
- **Minimum:** preserve unrecognized pragmas verbatim through `loads`/`unroll`/`dumps` instead of raising (a `preserve_pragmas=True` option, or a default no-op handler that re-emits the pragma unchanged).
- **Ideal:** recognize `braket verbatim` and pass the pragma plus its enclosing `box { ... }` through unmodified so downstream tooling handles them.

Also worth verifying as part of the fix: the enclosing `box` construct (`ast.Box`) round-trips through `unroll`/`dumps`, since verbatim requires both the pragma and the box.

ryanhill1:
## Summary of changes

Closes #340.

### Problem

Any program containing a `#pragma` was rejected with `ValidationError: Unsupported statement of type <class 'openqasm3.ast.Pragma'>`, because `ast.Pragma` had no entry in the visitor's dispatch map. That blocks verbatim execution on Braket QPUs through the qBraid stack: `#pragma braket verbatim` is the only remaining mechanism for fixed physical qubits, and pyqasm rejected the program client-side before it could be submitted.

Pragmas now pass through `loads`/`validate`/`unroll`/`dumps` unchanged.

### Verbatim boxes are not decomposed

Pass-through alone is not enough: unrolling still rewrote the gates inside the box, so a verbatim `prx(...) $1` came out as an `rz/rx/rz/rx/rz` chain the device cannot accept. A `braket verbatim` pragma now marks the `box` that immediately follows it, and gates inside that box take the existing external-gate path — validated and counted for depth/qubits, but emitted as written. Nested boxes inherit it; the marker is consumed by the next statement, so each verbatim box needs its own pragma.

The verbatim check is a literal match on the pragma command (`braket verbatim`). Every other pragma is opaque text that pyqasm copies through.

### `#pragma` vs `pragma`

`openqasm3`'s printer emits the bare `pragma` keyword. Both forms parse, but Braket emits and documents the hashed form, and the point of this change is round-tripping to Braket — so `Qasm3Module` now prints via a `Printer` subclass that writes `#pragma`. QASM 2 output is untouched (`Qasm2Module` still rejects pragmas via its statement whitelist).

### Limits of pass-through, documented

`src/README.md` gains a Pragmas section recording two things the design does not give you:

- **Qubit-renumbering passes do not rewrite pragma text.** `remove_idle_qubits()`, `reverse_qubit_order()` and `unroll(consolidate_qubits=True)` renumber the program but not the opaque pragma body, so `#pragma braket noise bit_flip(0.1) q[3]` can end up on a different qubit than it was written for, or outside the register — silently, since the output is still valid QASM. Inherent to pass-through; parsing vendor pragma grammar is not on the table.
- **A verbatim box should contain only device-native gates.** pyqasm does not enforce it, and a user-defined gate inside one is emitted as a call while unrolling drops its `gate` definition, so that output does not load back into pyqasm.

### Known gap: depth of verbatim gates

`#pragma braket verbatim box { crz(0.5) q[0], q[1]; }` emits one instruction but reports `depth == 12` — the depth of the decomposition that was deliberately skipped. Pre-existing on the `external_gates` path (`unroll(external_gates=["crz"])` reports the same on `main`) and tracked in #352; this PR makes it reachable without opting in per gate name. Only gates with a decomposition rule are affected: `verbatim box { rx; cz }` reports 2 and `verbatim box { ccx }` reports 1, both correct.

### Merge order

Merge after #344 and #345. Two caveats that would otherwise apply here are fixed by them: `unroll(consolidate_qubits=True)` on a verbatim box with physical qubits raises `AttributeError` on this branch alone and works merged (#344), and `remove_idle_qubits()` / `reverse_qubit_order()` correctly rewrite box bodies once merged (#345). Neither is on the path the issue needs (plain `unroll()` + `dumps()`).

## Tests

`tests/qasm3/test_pragma.py` covers round-tripping, `#pragma` normalisation, verbatim preservation, and the positional scoping in all four arrangements (pragma → gate → box, one pragma with two boxes, nested box inheritance, stacked pragmas). Validation inside a verbatim box is asserted to still reject undefined gates, wrong arity, duplicate operands and out-of-range indices — verbatim is not an escape hatch.

`tests/visualization/test_mpl_draw.py::test_draw_qasm3_pragma` covers the `Pragma` branch in the moment builder; `Pragma` subclasses neither `Statement` nor `QuantumStatement`, so without that branch a pragma reaches `raise ValueError(f"Unsupported statement: ...")`. Verified to fail when the branch is removed.


ryanhill1:
Reviewed all four findings; three were valid and are fixed in 993cd8b.

**Intervening pragma (major)** — valid. The condition excluded every `Pragma`, so a verbatim marker survived one. It is now cleared for any non-`Box` statement; because the clear happens before dispatch, a verbatim pragma still re-arms the flag for itself, and anything else — another pragma included — drops it. That matches the documented "the box that immediately follows it" rule. Regression test: `test_verbatim_marker_dropped_by_an_intervening_pragma`.

**Verbatim custom gate depth** — valid. Confirmed before the fix: a custom gate in a verbatim box was emitted as `my_gate(0.3) q[0], q[1];` but reported `depth == 3`, the depth of the body it does not emit; the same gate via `external_gates=["my_gate"]` reported 1. `_recording_ext_gate_depth` now also covers verbatim gates, so it reports 1. Pinned by `test_verbatim_custom_gate_counts_once_towards_depth`.

I did not add the preserve/restore of the previous state. The existing set-then-clear protocol already yields exactly one depth update for a nested custom gate (the inner call clears the flag, so the outer block skips); restoring the prior value would make both levels record, double-counting a single emitted gate. Verified: nested verbatim custom gate reports `depth == 1`, two sibling verbatim gates report 2.

**Both docstring findings** — valid, added.

One thing deliberately unchanged: a *basic* gate in a verbatim box still contributes its decomposition's depth (`prx` + `cz` reports 6). That is the pre-existing behaviour of the external-gate path — `unroll(external_gates=["prx"])` on the same program reports 6 as well — so it is not specific to verbatim and is out of scope here.

ryanhill1:
Valid, and reachable — fixed in b3bf100.

The text parser keeps pragmas global, but `loads()` also accepts a hand-built `ast.Program`, and that reaches the same visitor path. Confirmed before the fix: a box whose body ends with a verbatim pragma, followed by a second box with no pragma of its own, emitted `prx` undecomposed in the second box. `_visit_box_statement` now clears `_verbatim_pragma_pending` alongside restoring `_in_verbatim_box`, and `test_verbatim_marker_does_not_escape_a_box` builds exactly that program and pins the second box to its decomposed form.

TheGupta2012 (commented):
**Approved with comments** — well-designed feature, and the hard part (positional scoping) is right. None of the findings block merge; the first two are inherited rather than introduced here.

### Findings

- [ ] M1 — Verbatim gate with a decomposition rule reports decomposed depth (inherited)
- [ ] M2 — Custom gate in a verbatim box emits output that does not round-trip (inherited)
- [ ] M3 — Qubit-referencing pragmas silently desync from renumbering passes
- [ ] L1 — `mpl_draw` pragma branch is load-bearing but untested
- [ ] L2 — Docstring the honour-vs-forward asymmetry

### How it was tested

- Pragma round-trip through `loads`/`unroll`/`dumps`; bare `pragma` normalised to `#pragma`.
- Verbatim box preserves `rx`/`cz` as written (`depth=2`, `num_qubits=2`); plain `box { crz(0.5) q[0], q[1]; }` still decomposes to 12 statements, so no over-reach.
- **Scoping, all four positions:** pragma → gate → box leaves the box non-verbatim; one pragma with two boxes marks only the first; a nested box inherits; two stacked pragmas count only the immediately preceding one. `#pragma` inside a box is rejected by the parser itself (`pragmas must be global`), so the global-only assumption in `visit_statement` holds.
- **Validation is preserved inside a verbatim box:** undefined gate, wrong arity, duplicate operand and out-of-range index all still raise. Verbatim is not an escape hatch.
- QASM 2 still rejects pragmas via the statement whitelist.
- Suite on branch: **647 passed, 4 skipped**.
- Merged worktree of #341 + #344 + #345: **658 passing**; all six open PRs merge cleanly (`CHANGELOG.md` only) at **697 passing**.

### Next steps

Merge after #344 and #345 — both "Not fixed here" caveats are fixed by them, verified on a merged worktree, so that section can be deleted from the description rather than filed as follow-ups. Specifically: `consolidate_qubits=True` on a verbatim box with physical qubits raises `AttributeError` on this branch alone and works merged (#344); `remove_idle_qubits()`/`reverse_qubit_order()` correctly rewrite box bodies merged (#345). Then file the M1 follow-up and add the doc lines from M2/M3. L2 is attached as an applicable suggestion.

src/pyqasm/visitor.py
```diff
@@ -1257,8 +1263,9 @@ def _visit_custom_gate_operation(
         self._scope_manager.push_context(Context.GATE)
 
         # Pause recording the depth of new gates because we are processing the
-        # definition of a custom gate here - handle the depth separately afterwards
-        self._recording_ext_gate_depth = gate_name in self._external_gates
+        # definition of a custom gate here - handle the depth separately afterwards.
+        # A verbatim gate is emitted as written, so it counts once, like an external gate.
+        self._recording_ext_gate_depth = self._in_verbatim_box or gate_name in self._external_gates
```
TheGupta2012, line 1268: **[M1] Verbatim gate with a decomposition rule reports decomposed depth** — Implementation · Medium *(pre-existing; follow-up, not a blocker)*

**Rationale:** `#pragma braket verbatim box { crz(0.5) q[0], q[1]; }` emits exactly one instruction but reports `depth == 12` — the depth of the decomposition that was deliberately *not* performed. (`verbatim box { rx; cz }` → 2 and `verbatim box { ccx }` → 1 are both correct, so this is specific to gates that have a decomposition rule.)

Root cause: `_visit_external_gate_operation` calls `self._visit_basic_gate_operation(operation)` "just for validation" at `src/pyqasm/visitor.py:1351`, and that call still runs `_update_qubit_depth_for_gate`. The `_recording_ext_gate_depth` suppression this line extends guards only the *custom* gate path, not the basic-gate path.

This is pre-existing — on `main`, `unroll(external_gates=['crz'])` also reports `depth=12` while emitting one `crz` — so it is not a blocker here. It is worth flagging because this PR makes the defect reachable *without the user opting in per gate name*: any Braket verbatim program containing a decomposable gate now silently reports an inflated depth.

**Change requested:** file a follow-up against `_visit_external_gate_operation` (suppress depth recording around the validation-only call) and link it here. The "Not fixed here" section already sets the precedent.

src/pyqasm/visitor.py
```diff
@@ -1577,7 +1584,7 @@ def _visit_generic_gate_operation(  # pylint: disable=too-many-branches, too-man
         for _ in range(power_value):
             if isinstance(operation, qasm3_ast.QuantumPhase):
                 result.extend(self._visit_phase_operation(operation, inverse_value, ctrls))
-            elif operation.name.name in self._external_gates:
+            elif self._in_verbatim_box or operation.name.name in self._external_gates:
```
TheGupta2012, line 1587: **[M2] Custom gate in a verbatim box emits output that does not round-trip** — Design / Maintenance · Medium *(pre-existing in kind)*

**Rationale:** this branch makes *every* gate take the external path when inside a verbatim box, user-defined ones included. Given `gate mygate a { h a; }` and a verbatim box containing `mygate q[0];`, the emitted program keeps the `mygate q[0];` call while unrolling drops the `gate mygate` definition. Feeding that output back through pyqasm fails:

```
ValidationError: Unsupported / undeclared QASM operation: mygate
```

Pre-existing in kind — the same non-round-tripping output reproduces on `main` via `unroll(external_gates=['mygate'])`. The difference that matters: `external_gates` is an explicit per-name opt-in, whereas `_in_verbatim_box` makes every gate external implicitly, including user-defined ones a device cannot accept as native instructions. Braket verbatim boxes exist specifically for native hardware gates, so this combination is meaningless in the first place.

**Change requested:** either reject a non-native, non-basis custom gate inside a verbatim box with a clear error, or document that verbatim bodies must contain only device-native gates. At minimum a line in the docs added by this PR.

src/pyqasm/visitor.py
```diff
@@ -2934,6 +2941,40 @@ def _visit_delay_statement(
...
+
+        Args:
+            statement (qasm3_ast.Pragma): The pragma to inspect.
+
+        Returns:
+            bool: True for a 'braket verbatim' pragma, False for any other.
+        """
+        return statement.command.split() == ["braket", "verbatim"]
+
+    def _visit_pragma(self, statement: qasm3_ast.Pragma) -> list[qasm3_ast.Pragma]:
```
TheGupta2012, line 2956: **[M3] Qubit-referencing pragmas silently desync from renumbering passes** — Implementation · Medium

**Rationale:** pass-through means no transform can rewrite pragma text, so a pragma that names qubits by index goes stale as soon as a renumbering pass runs. Two verified cases:

```
qubit[4] q;
#pragma braket noise bit_flip(0.1) q[3]
h q[3];
```

After `unroll()` + `remove_idle_qubits()` the register shrinks to `qubit[1] q` and the gate is rewritten to `h q[0]`, but the pragma still reads `q[3]` — now pointing outside the declared register. Symmetrically, `reverse_qubit_order()` rewrites `h q[0]` → `h q[3]` while the pragma stays on `q[0]`, so the noise channel lands on a different qubit than the gate it was written for. Relatedly, `#pragma braket noise bit_flip(0.1) q[9]` against `qubit[1] q` passes `validate()` cleanly.

This is inherent to the pass-through design, and parsing vendor pragma grammar is not the suggestion. It is newly *reachable*, though — before this PR such programs were rejected outright — and it fails silently, producing a program that is valid QASM and wrong. Note this is distinct from the `box` recursion gap that #345 fixes: no amount of body-walking fixes opaque text.

**Change requested:** document in the docs added here that pragma bodies are opaque and are not rewritten by qubit-renumbering passes (`remove_idle_qubits`, `reverse_qubit_order`, `consolidate_qubits`), so pragmas referencing qubits by index may go stale. A warning when a renumbering pass runs on a module carrying pragmas would be a reasonable follow-up, not required here.

src/pyqasm/printer.py
```diff
@@ -134,6 +134,8 @@ def mpl_draw(  # pylint: disable=too-many-locals
     for s in program._statements:
         if isinstance(s, ast.QuantumPhase):
             global_phase += Qasm3ExprEvaluator.evaluate_expression(s.argument)[0]
+        elif isinstance(s, ast.Pragma):
+            continue  # pragmas carry no timing information
```
TheGupta2012, line 138: **[L1] This branch is load-bearing but untested** — Implementation · Low

**Rationale:** without this `continue`, the pragma reaches the moment builder and hits `raise ValueError(f"Unsupported statement: {statement}")` — `Pragma` subclasses neither `Statement` nor `QuantumStatement`, so it is not filtered downstream. Correct fix, but there is no pragma case anywhere in `tests/visualization/`, so a regression here would surface as a hard error for anyone drawing a Braket verbatim program.

**Change requested:** add one `mpl_draw` test over a program containing a pragma.

src/pyqasm/visitor.py
```diff
@@ -2934,6 +2941,40 @@ def _visit_delay_statement(
...
+        """
+        Visit a Pragma statement.
+
+        Pragmas carry vendor specific directives which pyqasm does not interpret, so they
+        are passed through unchanged. A 'braket verbatim' pragma additionally marks the box
+        that follows it, whose gates are then left as written instead of being decomposed.
```
TheGupta2012, line 2962: **[L2] Docstring the honour-vs-forward asymmetry** — Maintenance · Low

**Rationale:** `#pragma braket verbatim` followed by something other than a box is a no-op for pyqasm, but it is still copied to the output, where Braket would apply it to whatever follows. That is defensible given "the marker is consumed by the next statement", but the split between what pyqasm honours and what it merely forwards is currently implicit.

**Change requested:**

```suggestion
        Pragmas carry vendor specific directives which pyqasm does not interpret, so they
        are passed through unchanged. A 'braket verbatim' pragma additionally marks the box
        that follows it, whose gates are then left as written instead of being decomposed.
        A verbatim pragma not followed by a box is honoured by nothing in pyqasm but is
        still copied to the output, where the consumer applies it to whatever comes next.
```

Separately, in this file: the module-level `dumps` added in `modules/qasm3.py` shadows both `openqasm3.printer.dumps` and the public `pyqasm.dumps` (different signature). The shadowing is contained to that one file, so this is readability only — worth a glance, not a change request.

src/pyqasm/visitor.py
ryanhill1: Follow-up filed as #352 and linked from the description. No code change here, per your call that it is not worth growing the diff.

Reproduced both halves first: `verbatim box { crz(0.5) q[0], q[1]; }` reports `depth=12` while emitting one statement, and `unroll(external_gates=['crz'])` reports the same 12 on `main`. `verbatim box { rx; cz }` → 2 and `verbatim box { ccx }` → 1, so it is specific to gates with a decomposition rule as you said. #352 points at the validation-only `_visit_basic_gate_operation` call and the `_recording_ext_gate_depth` asymmetry.

The description now carries a "Known gap" section instead of leaving it unmentioned.

src/pyqasm/visitor.py
ryanhill1: Resolved in 456e03b — took the documentation option.

Confirmed the round-trip failure first: `gate mygate a { h a; }` plus a verbatim box containing `mygate q[0];` emits the call while unrolling drops the definition, and feeding that back gives `ValidationError: Unsupported / undeclared QASM operation: mygate`.

`src/README.md` now has a Pragmas section stating that a verbatim box should contain only device-native gates, that pyqasm does not enforce it, and what the output does if you ignore that. Went with documenting rather than rejecting since the combination is meaningless rather than dangerous, and rejecting would make pyqasm the arbiter of what a device considers native.

src/pyqasm/visitor.py
ryanhill1: Resolved in 456e03b — documented in the same Pragmas section.

Verified both directions on this branch:

```
remove_idle_qubits() : qubit[4] q -> qubit[1] q, h q[3] -> h q[0], pragma still q[3]
reverse_qubit_order(): h q[3] -> h q[0],          pragma still q[3]
```

The README now records that `remove_idle_qubits()`, `reverse_qubit_order()` and `unroll(consolidate_qubits=True)` renumber the program but not pragma text, that an index-referencing pragma can therefore land on the wrong qubit or outside the register, and that nothing raises because the output is still valid QASM. Left the warning-on-renumbering idea as a follow-up, per your note that it is not required here.

src/pyqasm/printer.py
ryanhill1: Resolved in 456e03b — `test_draw_qasm3_pragma` added to `tests/visualization/test_mpl_draw.py`.

Verified it is load-bearing: deleting the `continue` makes it fail with `ValueError: Unsupported statement: Pragma(...)`, exactly as you described.

One thing worth knowing about the motivation: a Braket verbatim program cannot be drawn today regardless of this branch — `Box` has no moment-builder case either, so `mpl_draw` on any boxed program raises `ValueError: Unsupported statement: Box(...)`, on `main` as well. So the test uses a bare pragma with plain gates to isolate the branch you flagged. Happy to file the `Box` drawing gap separately if you want it tracked.

src/pyqasm/visitor.py
ryanhill1: Resolved in 456e03b — suggestion applied verbatim to the `_visit_pragma` docstring.

Left the `dumps` shadowing in `modules/qasm3.py` alone, per your note that it is readability only and not a change request.

TheGupta2012 (approved):
**Approved.**

Verified against `main` now that #344 and #345 have landed: the branch merges with **no conflicts**, and both items in the "Not fixed here" section are closed by them —

- `unroll(consolidate_qubits=True)` on `#pragma braket verbatim box { rx(0.1) $0; cz $0, $1; }` succeeds (was `AttributeError: 'str' object has no attribute 'name'`) — #344.
- `remove_idle_qubits()` shrinks `qubit[4] q` to `qubit[2] q` and rewrites the box body; `reverse_qubit_order()` gives `rx(0.1) q[3]; cz q[3], q[2];` — #345.

Verbatim preservation is unaffected by either. Suite on the merged tree: 671 passed, 4 skipped.

The "Not fixed here" section can come out of the description now. M1 remains tracked in #352; M2/M3 documented in the new `src/README.md` Pragmas section, and the `mpl_draw` pragma branch is covered by `test_draw_qasm3_pragma` — confirmed load-bearing (removing the `continue` raises `ValueError: Unsupported statement: Pragma(...)`).

PR #346, opened 2026-08-04
typing and docstring fixes in modules folder and visitor.py

### LINKED ISSUE #347: Incorrect return types in visitor.py
A large number of functions in `visitor.py` erroneously claim in the docstring that they return `None` when they do not. This should be changed so their return types are properly documented in the docstring.

micpap25:
## Summary of changes
Normalize formatting of docstrings (periods, typos, etc.) and apply type hints to functions in `modules` folder and in `visitor`. 

Edit: Fixes #347


TheGupta2012:
Thanks for starting your contribution journey @micpap25 ! Changes lgtm, as they're mostly related to the docstrs. Can you please make sure that the CI passes and that the `CHANGELOG` is updated too?

micpap25:
@TheGupta2012 The reason that `mypy` was failing was because of the in-progress `Pragma` changes in the `Program` class at `ast.py`. I see there's an open PR regarding `Pragma`, so maybe I come back to this later?

micpap25:
@TheGupta2012 I think this PR is ready for merging, in retrospect this should have been part of the same PR as #348 since switching between them is somewhat impacting my workflow as they have a lot of overlap. Let me know what you think.

src/pyqasm/modules/base.py
```diff
@@ -22,21 +22,21 @@
 from abc import ABC, abstractmethod
 from collections import Counter
 from copy import deepcopy
-from typing import Optional
+from typing import Any, Callable, Optional, Self
```
ryanhill1: **B1 — blocking: breaks Python 3.10.**

`typing.Self` is 3.11+, but `pyproject.toml` sets `requires-python = ">=3.10"` and `main.yml` builds cp310 wheels on Linux, macOS x86_64/arm64 and Windows. Verified on a clean 3.10 env:

```
File ".../pyqasm/modules/base.py", line 25, in <module>
    from typing import Any, Callable, Optional, Self
ImportError: cannot import name 'Self' from 'typing'
```

`import pyqasm` fails outright — `from __future__ import annotations` doesn't help, since the import itself executes. It's used only for `copy() -> Self`; `-> QasmModule` works and matches the neighbouring methods.

Note the lint job pins 3.11, so this passes CI's format checks and only surfaces in the wheel matrix.

src/pyqasm/modules/qasm2.py
```diff
@@ -26,6 +26,7 @@
 from pyqasm.exceptions import ValidationError
 from pyqasm.modules.base import QasmModule
 from pyqasm.modules.qasm3 import Qasm3Module
+from pyqasm.visitor import QasmVisitor
```
ryanhill1: **B2 — blocking: `pylint` failure.**

```
src/pyqasm/modules/qasm2.py:29:0: W0611: Unused QasmVisitor imported from pyqasm.visitor
```

`accept(self, visitor)` below leaves the parameter unannotated, so `QasmVisitor` is referenced only from docstring text. Annotating `visitor: QasmVisitor` is presumably the intent and imports cleanly (`visitor.py` doesn't import `modules`); otherwise drop the import.

src/pyqasm/modules/qasm3.py
```diff
@@ -20,6 +20,7 @@
 from openqasm3.printer import dumps
 
 from pyqasm.modules.base import QasmModule
+from pyqasm.visitor import QasmVisitor
```
ryanhill1: **B3 — blocking: `pylint` failure.** Same as B2.

```
src/pyqasm/modules/qasm3.py:23:0: W0611: Unused QasmVisitor imported from pyqasm.visitor
```

src/pyqasm/visitor.py
```diff
@@ -2398,10 +2417,10 @@ def _visit_subroutine_definition(
            Reference: https://openqasm.com/language/subroutines.html#subroutines
 
         Args:
-            statement (qasm3_ast.SubroutineDefinition): The subroutine definition to visit.
+            statement (SubroutineDefinition | ExternDeclaration): The subroutine definition to visit.
```
ryanhill1: **B4 — blocking: `pylint` failure.**

```
src/pyqasm/visitor.py:2420:0: C0301: Line too long (101/100)
```

src/pyqasm/modules/base.py
```diff
@@ -742,28 +763,31 @@ def compare(self, other_module: QasmModule):
...
     @abstractmethod
-    def _qasm_ast_to_str(self, qasm_ast):
-        """Convert the qasm AST to a string"""
+    def _qasm_ast_to_str(self, qasm_ast: Program) -> str:
+        """Convert the qasm AST to a string."""
 
     @abstractmethod
-    def accept(self, visitor):
-        """Accept a visitor for the mßodule
+    def accept(self, visitor) -> None:
```
ryanhill1: **B2/B3 (cont.)** — `visitor` is unannotated here too. Annotating it in all three `accept` implementations resolves B2 and B3.

src/pyqasm/visitor.py
```diff
@@ -520,13 +525,16 @@ def _qubit_register_consolidation(
...
     ) -> None | qasm3_ast.Expression:
         """Handle function initialization expression.
 
         Args:
-            statement (Any): The statement to handle function initialization expression.
+            expression (Any): The statement to handle function initialization expression.
             init_value (Any): The value to handle function initialization expression.
+
+        Returns:
+            The resultant expression if
```
ryanhill1: **D1 — truncated sentence.** `The resultant expression if` just stops mid-clause.

Also, the `Args:` entry still reads `expression (Any)` while the signature was narrowed to `qasm3_ast.FunctionCall` — worth syncing. (The narrowing itself is correct; both call sites guard with `isinstance(..., FunctionCall)`.)

src/pyqasm/modules/base.py
```diff
@@ -346,16 +353,24 @@ def depth(self, decompose_native_gates=True):
...
+    def _remap_qubits(self, reg_name: str, size: int, idle_indices: list[int]) -> None:
         """Remap the qubits in a register after removing idle qubits and update the operations
-        using this register accordingly"""
+        using this register accordingly.
+        Args:
+            reg_name (str): The name of the register to be remapped.
+            size (int): The size of the register.
+            idle_indices (list[int]): Indices of idle qubits to be remapped away.
+
+            Returns:
```
ryanhill1: **D2 — malformed docstring.** `Returns:` is nested inside the `Args:` block with extra indentation, so it won't render as a section.

src/pyqasm/visitor.py
```diff
@@ -1063,11 +1073,12 @@ def _update_qubit_depth_for_gate(
...
         all_targets: list[list[Union[qasm3_ast.IndexedIdentifier, qasm3_ast.Identifier]]],
         ctrls: list[qasm3_ast.IndexedIdentifier],
-    ):
+    ) -> None:
         """Updates the depth of the circuit after applying a broadcasted gate.
 
         Args:
-            all_targes: The list of qubits on which a gate was just added.
+            all_targets: The list of qubits on which a gate was just added.
+            ctrls: THe list of control qubits for the gate.
```
ryanhill1: **D3 — typo.** `THe` → `The`.

src/pyqasm/visitor.py
```diff
@@ -2316,7 +2328,14 @@ def ravel(bit_ind):
...
 
     def _visit_forin_loop(self, statement: qasm3_ast.ForInLoop) -> list[qasm3_ast.Statement]:
-        # Compute loop variable values
+        """Visit a for-in loop statement element.
+
+        Args:
+            statement (qasm3_ast.ForInLoop): The for-in loop statement to visit.
+
+        Returns:
+            list[qasm3_ast.Statement]: The list of unrolled branch statements.
```
ryanhill1: **D4 — copy-paste error.** `_visit_forin_loop` documents "unrolled **branch** statements", lifted verbatim from `_visit_branching_statement`. Should be loop statements.

src/pyqasm/visitor.py
```diff
@@ -202,7 +202,7 @@ def _visit_quantum_register(
             register (QubitDeclaration): The register name and size.
 
         Returns:
-            None
+            The list containing the register
```
ryanhill1: **R1 — inaccurate.** Returns `[]` when `self._check_only`, so it isn't always "the list containing the register".

src/pyqasm/visitor.py
```diff
@@ -834,7 +843,7 @@ def _visit_barrier(  # pylint: disable=too-many-locals, too-many-branches
             statement (qasm3_ast.QuantumBarrier): The barrier statement to visit.
 
         Returns:
-            None
+            The list containing a single multi-qubit barrier statement.
```
ryanhill1: **R2 — inaccurate.** Same as R1: this method also has a `return []` path.

src/pyqasm/visitor.py
```diff
@@ -2398,10 +2417,10 @@ def _visit_subroutine_definition(
            Reference: https://openqasm.com/language/subroutines.html#subroutines
 
         Args:
-            statement (qasm3_ast.SubroutineDefinition): The subroutine definition to visit.
+            statement (SubroutineDefinition | ExternDeclaration): The subroutine definition to visit.
 
         Returns:
-            None
+            The list containing the statement if it is an ExternDeclaration, otherwise None.
```
ryanhill1: **R3 — inaccurate.** It returns `[]`, never `None`.

src/pyqasm/modules/base.py
```diff
@@ -630,10 +645,10 @@ def unroll(self, **kwargs):
             raise err
 
     @track_user_operation
-    def rebase(self, target_basis_set, in_place=True):
+    def rebase(self, target_basis_set: BasisSet, in_place: bool = True) -> QasmModule:
```
ryanhill1, line 690: **N1** — narrowing to `BasisSet` contradicts a tested contract: `tests/qasm3/test_rebase.py:266` deliberately calls `rebase("invalid_basis")` to exercise the `ValueError` path. Runtime is unaffected, but a type checker would now flag that test.

`target_basis_set` is also the one `Args:` entry left without a type, inconsistent with the rest of the PR.

ryanhill1 (commented):
Thanks for taking this on — the docstring/typing cleanup across `modules/` and `visitor.py` is a real improvement, and `mypy src`, `black`, `isort` and the test suite (648 passed, 3 skipped) are all green.

A few things need addressing before merge. Inline comments are labelled so you can reference them:

- **B1–B4 — blocking.** B1 breaks `import pyqasm` on Python 3.10; B2–B4 fail the `pylint` job in `format.yml`.
- **D1–D4** — malformed or truncated docstrings introduced here.
- **R1–R3** — documented return values that don't match the code.
- **N1–N7** — nits, take or leave.

**One more, no inline anchor:** the PR says *Fixes #347*, but two `Returns: None` cases in `visitor.py` are untouched and are exactly the class the issue describes:
- `_visit_include` (L3287) — returns `[include]` or `[]`
- `_visit_function_call` (L2465) — returns `tuple[Any | None, list[...]]`

Worth folding those in so the issue genuinely closes.

src/pyqasm/modules/base.py
```diff
@@ -22,21 +22,21 @@
...
-from pyqasm.elements import ClbitDepthNode, QubitDepthNode
+from pyqasm.elements import BasisSet, ClbitDepthNode, QubitDepthNode
 from pyqasm.exceptions import UnrollError, ValidationError
 from pyqasm.maps import QUANTUM_STATEMENTS
 from pyqasm.maps.decomposition_rules import DECOMPOSITION_RULES
 from pyqasm.visitor import QasmVisitor, ScopeManager
 
 
-def track_user_operation(func):
+def track_user_operation(func: Callable[..., Any]) -> Callable[..., Any]:
```
ryanhill1: **N2** — `Callable[..., Any] -> Callable[..., Any]` erases the signature of every decorated method (`depth`, `unroll`, `rebase`, `reverse_qubit_order`, `validate`, `remove_*`), so the return annotations added to them are invisible to callers. A `TypeVar("F", bound=Callable[..., Any])` preserves them.

src/pyqasm/modules/base.py
```diff
@@ -116,15 +116,15 @@ def num_qubits(self) -> int:
         return self._num_qubits
 
     @num_qubits.setter
-    def num_qubits(self, value: int):
-        """Setter for the number of qubits"""
+    def num_qubits(self, value: int) -> None:
+        """Setter for the number of qubits."""
         self._num_qubits = value
 
     def _add_qubit_register(self, reg_name: str, num_qubits: int):
```
ryanhill1: **N3** — missed: still unannotated, while its sibling `_add_classical_register` got `-> None`.

src/pyqasm/modules/base.py
```diff
@@ -567,7 +582,7 @@ def reverse_qubit_order(self, in_place=True):
         return qasm_module
 
     @track_user_operation
-    def validate(self):
+    def validate(self) -> None:
         """Validate the module"""
```
ryanhill1: **N4** — missed by the period pass: `"""Validate the module"""`.

src/pyqasm/modules/base.py
```diff
@@ -308,16 +315,16 @@ def remove_includes(self, in_place=True) -> Optional["QasmModule"]:
...
 
     @track_user_operation
-    def depth(self, decompose_native_gates=True):
+    def depth(self, decompose_native_gates: bool = True) -> int:
         """Calculate the depth of the unrolled openqasm program.
 
         Args:
-        decompose_native_gates (bool): If True, calculate depth after decomposing gates.
+            decompose_native_gates (bool): If True, calculate depth after decomposing gates.
                                 If False, treat all decompsable gates as a single gate operation.
```
ryanhill1: **N5** — the `Args:` indent was fixed above, but these continuation lines are still misaligned and `decompsable` is still misspelled.

src/pyqasm/modules/base.py
```diff
@@ -159,21 +159,28 @@ def _add_classical_register(self, reg_name: str, num_clbits: int):
...
         self._unrolled_ast = value
 
     def has_measurements(self) -> bool:
-        """Check if the module has any measurement operations."""
+        """Check if the module has any measurement operations.
+
+        Args:
+            None
+
+        Returns:
```
ryanhill1, line 229: **N6** — `Args: None` / `Returns: None` boilerplate is noise on a function whose signature already says so (same in `_get_gate_counts`, `compare`, and the three `accept` methods). It does match the existing `has_barriers`, so consistent either way — your call.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Types of changes:
 
 ### Improved / Modified
 - Consolidated the hardcoded `"__PYQASM_QUBITS__"` string literals scattered across `visitor.py`, `transformer.py` and `pulse/utils.py` into a single `INTERNAL_QUBIT_REGISTER` constant in `elements.py`, alongside an `is_internal_qubit_register()` helper that is now the one place the internal register is recognised. ([#325](https://github.com/qBraid/pyqasm/pull/325))
+- Added type hinting for `base.py`, `qasm2.py`, `qasm3.py`, and `visitor.py` docstrings. ([#346](https://github.com/qBraid/pyqasm/pull/346))
```
ryanhill1: **N7** — "type hinting for ... **docstrings**" isn't quite right; the hints went on signatures. Neither entry mentions #347, which is the substantive fix here.

CHANGELOG.md
```diff
@@ -35,6 +36,7 @@ Types of changes:
 - Fixed the `ctrl @` modifier not resolving gate aliases: `ctrl @ toffoli` / `ctrl @ ccnot` (aliases of `ccx`) and `ctrl @ cnot` / `ctrl @ CX` (aliases of `cx`) now escalate controls identically to their canonical gate instead of raising `Unsupported controlled QASM operation`. ([#320](https://github.com/qBraid/pyqasm/pull/320))
 - Fixed classical register declarations not being visible inside `box` scope, causing "Missing clbit register declaration" errors for measurement statements inside box blocks. ([#306](https://github.com/qBraid/pyqasm/pull/306))
 - Fixed the backend-dependent `dt` duration unit being incorrectly relabeled as `ns` when unrolling `delay` and `box` statements without a `device_cycle_time`. Since `dt` cannot be converted to SI units without a sample rate, it is now preserved as `dt`. ([#317](https://github.com/qBraid/pyqasm/pull/317))
+- Fixed grammatical typos in `base.py`, `qasm2.py`, `qasm3.py`, and `visitor.py` docstrings. ([#346](https://github.com/qBraid/pyqasm/pull/346))
```
ryanhill1: **N7 (cont.)** — this is a docs change, not a behavioural bug fix; both entries probably belong under *Improved / Modified*.

src/pyqasm/modules/base.py
micpap25: See the above comment as to why I do not do this; I think this can be done after the `Pragma` changes are implemented.

micpap25:
I believe I have covered everything addressed in the review, @ryanhill1 let me know what you think.

src/pyqasm/modules/base.py
```diff
@@ -22,21 +22,23 @@
...
 from pyqasm.exceptions import UnrollError, ValidationError
 from pyqasm.maps import QUANTUM_STATEMENTS
 from pyqasm.maps.decomposition_rules import DECOMPOSITION_RULES
 from pyqasm.visitor import QasmVisitor, ScopeManager
 
+F = TypeVar("F", bound=Callable[..., Any])
 
-def track_user_operation(func):
+
+def track_user_operation(func: F) -> F:
```
ryanhill1, line 96: **F1 — new mypy error, fails CI.**

```
src/pyqasm/modules/base.py:741: error: Argument 1 to "tabulate" has incompatible
  type "list[object]"; expected "Mapping[...] | Iterable[Iterable[Any]]"  [arg-type]
Found 1 error in 1 file (checked 37 source files)
```

The previous head passed `mypy src` clean, so this is new — but it's a *consequence of this fix being correct*. Now that `F -> F` preserves signatures through the decorator, `self.depth()` returns `int` instead of `Any`, so `table_data` in `compare()` infers as `list[object]` rather than `list[list[Any]]`.

Don't revert the TypeVar. Annotating the literal fixes it:

```python
table_data: list[list[Any]] = [
    ["Qubits", self.num_qubits, other_module.num_qubits],
    ...
]
```

src/pyqasm/visitor.py
```diff
@@ -520,13 +525,16 @@ def _qubit_register_consolidation(
...
     ) -> None | qasm3_ast.Expression:
         """Handle function initialization expression.
 
         Args:
-            statement (Any): The statement to handle function initialization expression.
+            expression (FunctionCall): The statement to handle function initialization expression.
             init_value (Any): The value to handle function initialization expression.
+
+        Returns:
+            None | Expression: The resultant expression if the expression is applied, otherwise None.
```
ryanhill1: **F2 — new `line-too-long`, fails CI.**

```
src/pyqasm/visitor.py:537:0: C0301: Line too long (101/100)
```

The completed D1 sentence runs one char over. Wrapping after `The resultant expression` does it.

src/pyqasm/visitor.py
```diff
@@ -735,10 +744,10 @@ def _visit_reset(self, statement: qasm3_ast.QuantumReset) -> list[qasm3_ast.Quan
         """Visit a reset statement element.
 
         Args:
-            statement (qasm3_ast.QuantumReset): The reset statement to visit.
+            statement (QuantumReset): The reset statement to visit.
 
         Returns:
-            None
+            list[QuantumReset]: The list of unrolled resets, or an empty list if self._check_only is True.
```
ryanhill1: **F2 (cont.)** — `visitor.py:750:0: C0301: Line too long (106/100)`.

Same shape as the others you wrapped onto a continuation line — this one didn't get wrapped.

src/pyqasm/visitor.py
```diff
@@ -2449,10 +2494,15 @@ def _visit_function_call(
...
-            statement (qasm3_ast.FunctionCall): The function call to visit.
-        Returns:
-            None
+            statement (FunctionCall): The function call to visit.
 
+        Returns:
+            tuple[Any | None, list[Statement | FunctionCall]]: If the function is undefined,
+                returns a tuple of None and an empty list.
+                If the function is defined, returns a tuple of the function's return
+                value and a list containing the fuction statement.
```
ryanhill1: **F3 — typo:** `fuction` → `function`.

src/pyqasm/visitor.py
```diff
@@ -2744,10 +2798,10 @@ def _visit_switch_statement(  # type: ignore[return]
         """Visit a switch statement element.
 
         Args:
-            statement (qasm3_ast.SwitchStatement): The switch statement to visit.
+            statement (SwitchStatement): The switch statement to visit.
 
         Returns:
-            list[qasm3_ast.Statement]: The list of statements generated by the switch statement.
+            list[Statement]: The list of statements generated by the switch statement.
```
ryanhill1: **F4 — missed by the `_check_only` pass.**

`_visit_switch_statement` reads `self._check_only`, but unlike the ~15 others you updated, its `Returns:` doesn't say so. Consistent with your *"Progress on updating all docstrings"* commit message — just flagging the two stragglers so the sweep is complete.

src/pyqasm/visitor.py
```diff
@@ -2850,10 +2910,12 @@ def _visit_delay_statement(
...
         """
         Visit a DelayInstruction statement.
+
         Args:
-            statement (qasm3_ast.DelayInstruction): The DelayInstruction statement to visit.
+            statement (DelayInstruction): The DelayInstruction statement to visit.
+
         Returns:
-            list[qasm3_ast.Statement]: The list of statements generated by the DelayInstruction.
+            list[Statement]: The list of statements generated by the DelayInstruction.
```
ryanhill1: **F4 (cont.)** — same for `_visit_delay_statement`.

src/pyqasm/modules/base.py
```diff
@@ -742,28 +745,28 @@ def compare(self, other_module: QasmModule):
...
     @abstractmethod
-    def _qasm_ast_to_str(self, qasm_ast):
-        """Convert the qasm AST to a string"""
+    def _qasm_ast_to_str(self, qasm_ast: Program) -> str:
+        """Convert the qasm AST to a string."""
 
     @abstractmethod
-    def accept(self, visitor):
-        """Accept a visitor for the mßodule
+    def accept(self, visitor) -> None:
```
ryanhill1: **B2/B3 (cont.) — leaving this one open, non-blocking.**

Dropping the unused imports resolved the pylint errors, so B2 and B3 are closed. But `visitor` is still unannotated here and in `qasm2.py:99` / `qasm3.py:45` — the last untyped signature in the files this PR touches.

`base.py` already imports `QasmVisitor` legitimately (L579, L627), so annotating is free here; the other two would need the import back. Entirely your call whether it's in scope.

ryanhill1 (commented):
Re-checked everything at `bdd18e6`. **18 of the 20 threads are genuinely fixed** — I've resolved those. Verified rather than eyeballed:

- **B1** — built a real 3.10 env: `import pyqasm` now succeeds. The `TypeVar` swap is the right call.
- **#347** — my AST sweep over `visitor.py` for documented-`None`-but-returns-something now comes back **empty**. `_visit_include` and `_visit_function_call` were the two I'd flagged; both are done. The issue genuinely closes.
- **R1–R3** — the new *"or an empty list if `self._check_only` is True"* phrasing is accurate against the code on all three.
- **Tests: 648 passed, 3 skipped.** `black` and `isort` clean.

Nice work — that's the substance of the review dealt with, including the blocker.

**Four small things remain, two of which will red-X the `format` job**, so this isn't quite mergeable yet. All inline below.

| | |
|---|---|
| **F1** | mypy regression in `base.py` — new, and a side-effect of your N2 fix |
| **F2** | two new `line-too-long` in `visitor.py` |
| **F3** | new typo, `fuction` |
| **F4** | two docstrings not yet carrying the `_check_only` note |

Left open deliberately: **N1** (the `BasisSet` narrowing vs. `test_rebase.py:266`) — you added the type to the docstring but kept the annotation, which is a fair call on a nit; and **B2/B3 (cont.)**, since `accept(self, visitor)` is still unannotated in all three modules. Neither blocks.

src/pyqasm/modules/base.py
micpap25: Leaving this out of scope since it would raise `pylint` issues from `Pragma`.

micpap25:
Tossed in one more fix, while looking at `_visit_switch_statement` I noticed there was a `pylint` comment that wasn't doing anything so I moved it to where it should be.

src/pyqasm/visitor.py
```diff
@@ -2738,16 +2794,17 @@ def _visit_alias_statement(self, statement: qasm3_ast.AliasStatement) -> list[No
 
         return []
 
-    def _visit_switch_statement(  # type: ignore[return]
+    def _visit_switch_statement(
         self, statement: qasm3_ast.SwitchStatement
-    ) -> list[qasm3_ast.Statement]:
+    ) -> list[qasm3_ast.Statement]:  # type: ignore[return]
```
ryanhill1: **H1 — blocking: this relocation breaks mypy.**

```
src/pyqasm/visitor.py:2797: error: Missing return statement  [return]
```

The comment wasn't orphaned. For a **multi-line** signature, mypy matches `# type: ignore` against the *first* line of the definition — the `def` line — even though it *reports* `Missing return statement` at 2797. That asymmetry is why it looked misplaced.

Verified both directions on this branch:

| `# type: ignore[return]` on | `mypy src` |
|---|---|
| `) -> list[...]:` (2799, current) | 1 error |
| `def _visit_switch_statement(` (2797, previous) | **Success: no issues found in 37 source files** |

Simplest fix is to put it back where it was. If it bothers you as a lint-suppression, the alternative is to drop the ignore entirely and give the function an explicit fallthrough `return []` — mypy is right that the `if`/`elif` chain over the cases has no exhaustive final branch. That's a behaviour change though, so it belongs in its own PR rather than this one.

ryanhill1 (commented):
Re-checked at `576fcb0a`. **F1-F4 are all fixed** — resolved those threads.

- **F1** — `table_data: list[list[Any]]` is exactly right; keeps the `TypeVar` win from N2 and silences the inference.
- **F2** — both lines wrapped; `pylint` on the four touched files is back to **10.00/10**.
- **F3** — `fuction` fixed.
- **F4** — `_visit_switch_statement` and `_visit_delay_statement` both carry the `_check_only` note now. My consistency sweep over `visitor.py` finds **0** methods left that read `_check_only` without documenting it, and **0** docstring/signature return mismatches across all four files.

**648 passed, 3 skipped. pylint 10.00/10, black, isort clean. `import pyqasm` verified on a real 3.10 env.**

### One regression, from the last commit

`576fcb0a` ("orphaned type comment") moved the `# type: ignore[return]` off the `def` line, which breaks mypy:

```
src/pyqasm/visitor.py:2797: error: Missing return statement  [return]
Found 1 error in 1 file (checked 37 source files)
```

The comment wasn't orphaned — its placement was load-bearing. Detail inline. **This is the only thing failing CI**; put it back and the PR is green everywhere.

Still open, neither blocking, both already discussed: **N1** (`BasisSet` narrowing) and `accept(self, visitor)` being unannotated in the three modules.

micpap25:
@ryanhill1 Should be done, fixed B2 / B3.

src/pyqasm/visitor.py
```diff
@@ -3337,10 +3411,10 @@ def visit_statement(
         """Visit a statement element.
 
         Args:
-            statement (qasm3_ast.Statement | qasm3_ast.Pragma): The statement to visit.
+            statement (qasm3_ast.Statement): The statement to visit.
```
ryanhill1: **J1 — new with the latest commits.** The signature is `statement: qasm3_ast.Statement | qasm3_ast.Pragma` — the previous doc had it right. Restore `| Pragma`.

src/pyqasm/visitor.py
```diff
@@ -3373,27 +3447,27 @@ def visit_basic_block(
         """Visit a basic block of statements.
 
         Args:
-            stmt_list (Sequence[qasm3_ast.Statement | qasm3_ast.Pragma]): The statements to visit.
+            stmt_list (list[qasm3_ast.Statement]): The list of statements to visit.
```
ryanhill1: **J2 — new with the latest commits.** Signature is `Sequence[qasm3_ast.Statement | qasm3_ast.Pragma]`; the doc now claims `list[qasm3_ast.Statement]`. Restore `| Pragma` (and `Sequence` while you're here).

src/pyqasm/modules/base.py
```diff
@@ -646,10 +652,10 @@ def unroll(self, **kwargs):
                 unroll_barriers (bool): If True, barriers will be unrolled. Defaults to True.
                 max_loop_iters (int): Max number of iterations for unrolling loops. Defaults to 1e9.
                 check_only (bool): If True, only check the program without executing it.
-                                   Defaults to False.
-                device_qubits (int): Number of physical qubits available on the target device.
-                consolidate_qubits (bool): If True, consolidate all quantum registers into
-                                           single register.
+                Defaults to False.
+            device_qubits (int): Number of physical qubits available on the target device.
```
ryanhill1: **J3 — missed in my earlier rounds.** `device_qubits` is dedented to the `**kwargs` level (reads as a separate parameter) and `consolidate_qubits` is nested beneath it. Note open #349 edits this exact block — it removes `device_qubits` from this docstring entirely (it's a `loads()` kwarg). Simplest here: re-indent `consolidate_qubits` back to the same level as the other kwargs and leave `device_qubits` to #349.

src/pyqasm/visitor.py
```diff
@@ -806,7 +817,15 @@ def _expand_barrier_ranges(
     ) -> list:
         """Replace RangeDefinition-containing qubits in a barrier with their
         expanded IndexedIdentifier equivalents so that consolidate_qubit_registers
-        only sees IntegerLiteral indices."""
+        only sees IntegerLiteral indices.
+
+        Args:
+            barrier (QuantumBarrier): The barrier with qubits to consolidate.
+            barrier_qubits (QuantumBarrier): The qubits to replace.
```
ryanhill1: **J4 — missed in my earlier rounds.** Wrong type: `barrier_qubits` is `list[qasm3_ast.IndexedIdentifier | qasm3_ast.Identifier]`, not `QuantumBarrier`. "The qubits to replace" also undersells it — these are the expanded operands substituted for ranges.

ryanhill1 (commented):
Re-checked at `cdedb54`. Toolchain on this head: **727 passed, 3 skipped; `black`, `isort` clean; `pylint` 10.00/10; `mypy` clean** (the only report is the pre-existing `tabulate` stubs gap, identical on `main`). `# type: ignore[return]` verified back on the `def` line — mypy no longer errors.

### Resolved this round

- **H1** — the ignore comment is back where it's load-bearing; mypy green. Resolving the thread.
- **B2/B3 (cont.)** — all three `accept` implementations now take `visitor: QasmVisitor`, imported via `modules.base`, so pylint stays clean. Resolving the thread.

### New in the latest commits (J1–J2)

Both docstrings were *more* accurate before the last edits — the signatures still (correctly) accept `Pragma`:

| | |
|---|---|
| **J1** | `visit_statement` Args doc dropped `\| Pragma` |
| **J2** | `visit_basic_block` Args doc dropped `\| Pragma` and says `list` for a `Sequence` |

### Missed in my earlier rounds (J3–J5) — not new, flagging for completeness

| | |
|---|---|
| **J3** | `unroll()` kwargs docstring: `device_qubits` dedented out of the `**kwargs` nesting, `consolidate_qubits` nested under it. Heads-up: open #349 edits this same block (it removes `device_qubits`, which is a `loads()` kwarg, not an `unroll()` one) — coordinate to avoid a conflict |
| **J4** | `_expand_barrier_ranges` Args types `barrier_qubits` as `QuantumBarrier`; it's `list[IndexedIdentifier \| Identifier]` |
| **J5** | `_visit_branching_statement`: double comma |

### Still open, unchanged, non-blocking

- **N1** — `BasisSet` narrowing vs `test_rebase.py:266`; you kept the annotation and documented the type, which I already called a fair resolution. Leaving as-is.

All five J items are one-line docstring fixes — none touch behaviour. Fix those and this is mergeable from my side.

src/pyqasm/visitor.py
```diff
@@ -2212,10 +2247,11 @@ def _visit_branching_statement(
         """Visit a branching statement element.
 
         Args:
-            statement (qasm3_ast.BranchingStatement): The branching statement to visit.
+            statement (BranchingStatement): The branching statement to visit.
 
         Returns:
-            None
+            list[Statement]: The list of unrolled branch statements, , or an empty list
```
ryanhill1: **J5 — missed in my earlier rounds.** Double comma: `statements, , or an empty list`.

src/pyqasm/modules/base.py
ryanhill1: **Update: #349 has merged, which turns this into a real merge conflict** (the PR is now `CONFLICTING` against `main`) — and it changes the guidance above.

`main`'s version of this block now reads:

```
check_only (bool): If True, only check the program without executing it.
                   Defaults to False.
consolidate_qubits (bool): If True, consolidate all quantum registers into
                           single register.
```

i.e. `device_qubits` is gone entirely (it's a `loads()` kwarg, not an `unroll()` one) and the indentation is already correct.

So when you merge `main` into your branch, resolve this hunk by **taking `main`'s docstring body wholesale** — that fixes J3 for free. Keep your side's signature (`def unroll(self, **kwargs: Any) -> None:`), which `main` doesn't have.

micpap25: To keep consistent with everything else in `visitor.py` I am reducing the intents on the lines `Defaults to False.` and `single register.` so they have a single indent over their previous lines.

src/pyqasm/pulse/visitor.py
```diff
@@ -754,11 +754,13 @@ def _visit_function_call(  # pylint: disable=too-many-branches, too-many-stateme
 
         return _return_value, [statement]
 
-    def visit_statement(self, statement: qasm3_ast.Statement) -> list[qasm3_ast.Statement]:
+    def visit_statement(
+            self, statement: qasm3_ast.Statement | qasm3_ast.Pragma
```
ryanhill1: **K2 — black failure, fails CI.**

```python
    def visit_statement(
            self, statement: qasm3_ast.Statement | qasm3_ast.Pragma
        ) -> list[qasm3_ast.Statement]:
```

black wants the standard continuation indent:

```python
    def visit_statement(
        self, statement: qasm3_ast.Statement | qasm3_ast.Pragma
    ) -> list[qasm3_ast.Statement]:
```

`black src` fixes it. (Also note this file is `OpenPulseVisitor` — see K1 in the review body: the J1/J2 targets were the same-named methods in `src/pyqasm/visitor.py`.)

src/pyqasm/visitor.py
ryanhill1: **K1 (= J1, still open)** — this is the method J1 pointed at: signature says `qasm3_ast.Statement | qasm3_ast.Pragma`, Args doc still says `statement (qasm3_ast.Statement)`. The `J1, J2` commit edited the same-named method on `OpenPulseVisitor` instead.

src/pyqasm/visitor.py
ryanhill1: **K1 (= J2, still open)** — same here: `stmt_list` is `Sequence[qasm3_ast.Statement | qasm3_ast.Pragma]`, doc says `list[qasm3_ast.Statement]`.

ryanhill1 (commented):
Re-checked at `327e1a7`. **J3, J4, J5 are done** — resolved those threads. J3 resolved itself the nice way: #349 merged into `main` and your merge brought its version of the `unroll()` docstring through, so `device_qubits` is gone and `consolidate_qubits` sits at the right level. J4 and J5 are exactly right.

Two things still stand between this and green CI:

**K1 — J1/J2 landed in the wrong file.** The `J1, J2` commit edited `OpenPulseVisitor.visit_statement` / `visit_basic_block` in `src/pyqasm/pulse/visitor.py` — same method names, different class. The J1/J2 comments were anchored in `src/pyqasm/visitor.py` (`QasmVisitor`, L3414 / L3450), where the docstrings still say `Statement` without `| Pragma`. Easy mix-up given the twin names; the original threads stay open.

**K2 — the pulse-visitor edit fails `black`**, so the format job is red:

```
would reformat src/pyqasm/pulse/visitor.py
```

The continuation lines got 12/8-space indents where black wants 8/4 (diff inline). Running `black src` fixes it in place.

On the pulse-visitor change itself: no objection to keeping it — the signature widening mirrors the `QasmVisitor` one and `mypy` is clean either way. But it wasn't part of any review ask, so if you'd rather keep this PR tight, reverting the two signature changes there and keeping only docstring edits is also fine.

Everything else on this head: **696 passed, 3 skipped; pylint 10.00/10; mypy clean; isort clean.** Once J1/J2 land in `visitor.py` and black is happy, this is mergeable from my side.

micpap25:
@ryanhill1 Sorry about how many times this has gone back-and-forth, I think this should resolve everything.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Types of changes:
 
 ### Improved / Modified
 - Consolidated the hardcoded `"__PYQASM_QUBITS__"` string literals scattered across `visitor.py`, `transformer.py` and `pulse/utils.py` into a single `INTERNAL_QUBIT_REGISTER` constant in `elements.py`, alongside an `is_internal_qubit_register()` helper that is now the one place the internal register is recognised. ([#325](https://github.com/qBraid/pyqasm/pull/325))
+- Added / updated type hinting for `base.py`, `qasm2.py`, `qasm3.py`, `visitor.py`, and `pulse/visitor.py` signatures. Fixed grammatical typos in `base.py`, `qasm2.py`, `qasm3.py`, and `visitor.py` docstrings. Fixes incorrect return types in the docstrings of `visitor.py`. ([#346](https://github.com/qBraid/pyqasm/pull/346))
```
ryanhill1: The PR touches two methods in `pulse/visitor.py` (`visit_statement` / `visit_basic_block`), so naming the whole file alongside the four fully-swept modules overstates it a bit:

```suggestion
- Added / updated type hinting for `base.py`, `qasm2.py`, `qasm3.py`, and `visitor.py` signatures, plus the `visit_statement` / `visit_basic_block` signatures in `pulse/visitor.py`. Fixed grammatical typos in `base.py`, `qasm2.py`, `qasm3.py`, and `visitor.py` docstrings. Fixes incorrect return types in the docstrings of `visitor.py`. ([#346](https://github.com/qBraid/pyqasm/pull/346))
```

ryanhill1 (commented):
Re-checked at `f5904b9`. **K1 and K2 both fixed** — the `| Pragma` docs are in the right file now and black is happy. Resolved J1/J2/K1/K2.

Everything is green on this head for the first time: **696 passed, 3 skipped; black, isort, pylint 10.00/10, mypy all clean; py3.10 import verified.**

One last small thing, inline: the updated CHANGELOG entry slightly oversells the `pulse/visitor.py` coverage. Take the suggestion and this is good to go from my side.

src/pyqasm/pulse/visitor.py
```diff
@@ -754,11 +754,13 @@ def _visit_function_call(  # pylint: disable=too-many-branches, too-many-stateme
...
+        self, statement: qasm3_ast.Statement | qasm3_ast.Pragma
+    ) -> list[qasm3_ast.Statement]:
         """Visit a statement element.
 
         Args:
-            statement (qasm3_ast.Statement): The statement to visit.
+            statement (Statement | Pragma): The statement to visit.
 
         Returns:
             None
```
ryanhill1: Nit, non-blocking either way: the method you just edited still says `Returns: None` against a `list[Statement]` signature — the same class of mismatch this PR fixed across `visitor.py` for #347. Since you're in the file anyway:

```suggestion
            list[Statement]: The list of resulting statements.
```

(`pulse/visitor.py` has three more of these in untouched methods — `_visit_barrier`, `_visit_classical_declaration`, `_visit_function_call` — but those can ride with a follow-up issue rather than this PR.)

ryanhill1 (approved):

Issue #188, opened 2025-05-24
Reduce LOC in `visitor.py`


TheGupta2012:
#### **Feature Description**
Refactor the `visitor.py` module to reduce its overall lines of code (LOC) without sacrificing readability or maintainability. The goal is to streamline the visitor logic, making it more concise and easier to maintain.

#### **Implementation**
- **Identify Repetitive Patterns:** Review visitor methods for similar logic that can be abstracted into helper functions or base classes.
- **Leverage Pythonic Constructs:** Use decorators, context managers, or utility functions to handle common pre/post-processing tasks.
- **Parameterize Common Operations:** Where multiple methods differ only by a node type or minor detail, parameterize the differences and use a single generalized function.
- **Inline Simple Methods:** For methods that only call a superclass or perform a trivial transformation, consider inlining or removing them.
- **Document Changes:** Ensure all refactorings are well-documented and covered by existing or new tests to prevent regressions.

---

> **Note:**  
> All refactoring should preserve existing functionality and pass the current test suite. If any behavioral changes are required, they should be proposed and discussed separately.

micpap25:
@TheGupta2012 I am interested in working on this issue alongside #347

PR #348, opened 2026-08-05
visitor.py LOC cleanup

micpap25:
## Summary of changes
Progress on #188 
Removes the `_handle_function_init_expression` which can be made inline, and adds the `check_only_return_empty` decorator for functions which can be given simple boiler-plate logic for the `self._check_only` parameter.


src/pyqasm/visitor.py
```diff
@@ -1601,11 +1592,27 @@ def _visit_generic_gate_operation(  # pylint: disable=too-many-branches, too-man
                 ),
             )
 
-        if self._check_only:
-            return []
-
         return result
 
+    def _validate_duration_or_stretch_statements(
```
TheGupta2012: Don't think this refactor is necessary here as its just calling the two functions inside, might be better to think in different terms

src/pyqasm/visitor.py
```diff
@@ -696,57 +711,42 @@ def _visit_measurement(  # pylint: disable=too-many-locals,too-many-branches,too
                 ),
             )
 
-        if self._check_only:
-            return []
-
         return unrolled_measurements
 
-    def _resolve_unindexed_reset_qubit(self, statement: qasm3_ast.QuantumReset) -> bool:
```
TheGupta2012, line 759: Why not just update this method's definition itself? Don't think that refactoring it inside the visit reset does a lot of changes.

src/pyqasm/visitor.py
```diff
@@ -83,6 +84,20 @@
 logger = logging.getLogger(__name__)
 logger.propagate = False
 
+def easy_check_only(func):
```
TheGupta2012: This is a nice addition! We want to find and refactor such lines in our original code. However, renaming to `semantic_check_gate` should be better

TheGupta2012 (commented):
Thanks @micpap25 for working on this! Can you please resolve the comments and continue the work for refactoring the visitor? This seems like a good start!

src/pyqasm/visitor.py
```diff
@@ -84,6 +85,21 @@
...
 
 
+def semantic_check_gate(func):
+    """Decorator for functions which use check_only to return an empty list."""
+
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        """Wrapper that intercepts the return value and replaces it with an empty list."""
+        result = func(self, *args, **kwargs)
+        if self.check_only:
```
ryanhill1: **B1 — blocking: wrong attribute name, breaks 539 of 648 tests.**

The attribute is `self._check_only` (set at L144); there is no `check_only` property on `QasmVisitor`.

```
src/pyqasm/visitor.py:3260: in visit_statement
E   AttributeError: 'QasmVisitor' object has no attribute 'check_only'
src/pyqasm/visitor.py:95: AttributeError
```

Every decorated method raises on first call. `s/self.check_only/self._check_only/` alone takes the suite from 539 failed to 6 failed.

src/pyqasm/visitor.py
```diff
@@ -2780,6 +2755,7 @@ def _visit_switch_statement(  # type: ignore[return]
         #    each element in the list of the values
         #    should be of const int type and no duplicates should be present
 
+        @semantic_check_gate
```
ryanhill1: **B2 — blocking: decorator applied to a nested function, breaks all 6 switch tests.**

`_evaluate_case(statements)` is a closure, not a method. The wrapper's first positional parameter is named `self`, so it binds to the `statements` list:

```
tests/qasm3/test_switch.py::test_switch
E   AttributeError: 'list' object has no attribute '_check_only'
src/pyqasm/visitor.py:95: AttributeError
```

(That traceback is after B1 is fixed — this is an independent bug.) The closure already reads `self._check_only` from the enclosing scope, so the decorator buys nothing here. Removing it is the fix; with B1 and B2 both fixed the suite is 648 passed / 3 skipped.

src/pyqasm/visitor.py
```diff
@@ -1086,6 +1082,7 @@ def _update_qubit_depth_for_gate(
                     qubit_node = self._module._qubit_depths[(qubit_name, qubit_id)]
                     qubit_node.depth = max_involved_depth
 
+    @semantic_check_gate
```
ryanhill1: **B3 — blocking: decorator displaces a block-scoped `# pylint: disable`, adding 10 pylint errors.**

A standalone `# pylint: disable=` comment is block-scoped. Inserting the decorator above it detaches it from the class body scope, so the suppression stops applying:

```
visitor.py:1203:4: R0914: Too many local variables (17/15)
visitor.py:1437:4: R0914: Too many local variables (20/15)
visitor.py:1950:4: R0912: Too many branches (25/12)
... 10 total
```

`main` scores 10.00/10 on this file. Moving the decorator *below* the comment at both sites (here and L1714) restores it — verified, drops pylint to a single remaining error (B4).

ryanhill1 (commented):
Thanks for picking up #188 — consolidating the repeated `if self._check_only: return []` tail into a decorator is the right instinct, and 26 net lines out of `visitor.py` is real progress. The `_handle_function_init_expression` inlining is behaviour-preserving as far as I can tell, and `mypy`, `black` and `isort` are all clean.

The decorator itself has two bugs that need fixing before this can go further, and once they're fixed there's a design question about early returns that I think decides whether this approach works at all.

**Test suite on the branch as pushed: 539 failed, 109 passed** (`pytest tests --deselect tests/cli`).

### What I ran

Patching B1 alone → 6 failures left. Patching B1 + B2 → **648 passed, 3 skipped**, matching `main`. So B1 and B2 account for the entire breakage, and the rest of the refactor doesn't regress anything the suite covers.

Reordering the two displaced `# pylint: disable` comments (B3) drops pylint from 11 errors to 1. `main` scores 10.00/10 on `visitor.py`, so all 11 are new here.

### The design question — R1

The decorator wraps the *whole* function, so it now also swallows **early** returns that previously bypassed the `_check_only` gate. Three are reachable, all on openpulse paths. Demonstrated with a pulse program:

```python
m = loads(pulse_program_with_defcal_measure)
m.validate()
len(m._unrolled_ast.statements)
# main: 4  (QubitDeclaration, CalibrationGrammarDeclaration, CalibrationStatement, QuantumMeasurementStatement)
# PR:   3  (QuantumMeasurementStatement dropped)
```

No test covers this, which is why the suite is green once B1/B2 are fixed. It may well be that returning `[]` there is *more* correct — but #188 says explicitly that behavioural changes should be proposed separately, so this needs to be either a deliberate, tested decision or excluded from the refactor. Details inline at R1.

### Not blocking, worth knowing

- **Merge conflict with #346.** That PR modifies `_handle_function_init_expression`; this one deletes it. Whichever lands second will need a rebase.
- **No CHANGELOG entry**, and you've flagged tests as outstanding yourself — #188 asks for coverage on refactors specifically to catch things like R1.

Labels: **B** blocking, **R** semantics, **D** dead code, **N** nits.

src/pyqasm/visitor.py
```diff
@@ -1715,17 +1703,15 @@ def _visit_constant_declaration(
...
+            if function_name in FUNCTION_MAP and isinstance(init_value, (float, int)):
+                statement.init_expression = qasm3_ast.FloatLiteral(init_value)
 
-        if self._check_only:
-            return []
+        self._handle_extern_function_cleanup(statements, statement)
 
         return statements
 
+    @semantic_check_gate
```
ryanhill1: **B3 (cont.)** — same displacement here, for `# pylint: disable=too-many-branches, too-many-statements, too-many-locals`.

src/pyqasm/visitor.py
```diff
@@ -84,6 +85,21 @@
...
+def semantic_check_gate(func):
+    """Decorator for functions which use check_only to return an empty list."""
+
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        """Wrapper that intercepts the return value and replaces it with an empty list."""
+        result = func(self, *args, **kwargs)
+        if self.check_only:
+            return []
+        else:
```
ryanhill1: **B4 — blocking: `pylint` failure.**

```
visitor.py:95:8: R1705: Unnecessary "else" after "return" (no-else-return)
```

```python
if self._check_only:
    return []
return result
```

src/pyqasm/visitor.py
```diff
@@ -558,6 +557,7 @@ def _validate_bitstring_literal_width(self, init_value, base_size, var_name, sta
                 span=statement.span,
             )
 
+    @semantic_check_gate
```
ryanhill1: **R1 — semantics: this decorator also gates an early return that previously bypassed the check.**

`_visit_measurement` returns `[statement]` at L607 on the pulse-gate path, before the old `if self._check_only: return []` was ever reached. Under the decorator that early return becomes `[]`.

Measured on a `defcalgrammar "openpulse"` program with a `defcal measure $0`: `validate()` leaves 4 statements in `_unrolled_ast` on `main` and 3 here — the `QuantumMeasurementStatement` is dropped.

The three reachable cases are this one, `_visit_reset` (L744) and `_visit_generic_gate_operation` (L1477). The other three early returns I checked are safe: `_visit_phase_operation` L1400 and `_visit_classical_declaration` L1828 delegate to already-gated methods, and `_visit_classical_assignment` L1977 returns `[]` regardless.

If the new behaviour is what you want, it needs tests and a note — it's a behavioural change, not a refactor. If not, the decorator can't be applied to methods with early returns.

src/pyqasm/visitor.py
```diff
@@ -731,22 +728,24 @@ def _resolve_unindexed_reset_qubit(self, statement: qasm3_ast.QuantumReset) -> b
 
         return is_internal_qubit_register(qubit_name)
 
+    @semantic_check_gate
```
ryanhill1: **R1 (cont.)** — `_visit_reset` returns `[statement]` at L744 on the `_resolve_unindexed_reset_qubit` (internal qubit register) path, now gated to `[]`.

src/pyqasm/visitor.py
```diff
@@ -1439,11 +1431,9 @@ def _visit_phase_operation(
 
         # if it were in function scope, then the args would have been evaluated and added to the
         # qubit list
-        if self._check_only:
-            return []
-
         return [operation]
 
+    @semantic_check_gate
```
ryanhill1: **R1 (cont.)** — `_visit_generic_gate_operation` returns `stmts` at L1477 from `PulseUtils.process_qubits_for_openpulse_gate`, now gated to `[]`.

src/pyqasm/visitor.py
```diff
@@ -1895,7 +1880,7 @@ def _visit_classical_declaration(
             angle_bit_string=angle_val_bit_string,
         )
 
-        if isinstance(base_type, qasm3_ast.DurationType):
+        if isinstance(base_type, (qasm3_ast.DurationType, qasm3_ast.StretchType)):
```
ryanhill1: **D1 — dead code: this `StretchType` branch is unreachable.**

`_visit_classical_declaration` raises at L1761-1768 whenever a `StretchType` declaration has an `init_expression`. If it doesn't have one, `init_value` stays `None`, and `validate_duration_literal_value` no-ops on its `duration_literal is not None` guard.

So neither branch of the stretch path can reach the new check. I couldn't construct a program where it changes anything — `stretch s = 0ns;` and `stretch s = -5ns;` raise the same "Assignment to 'stretch' type variable is not allowed" on both `main` and this branch.

Also worth noting this is a *functional* change smuggled into a LOC-reduction PR, which #188 asks be proposed separately. If enabling stretch value validation is the goal, it needs the earlier guards relaxed first, plus tests.

src/pyqasm/visitor.py
```diff
@@ -2083,7 +2065,7 @@ def _visit_classical_assignment(
             )
             rvalue_eval = rvalue_raw
 
-        if isinstance(lvar_base_type, qasm3_ast.DurationType):
+        if isinstance(lvar_base_type, (qasm3_ast.DurationType, qasm3_ast.StretchType)):
```
ryanhill1: **D1 (cont.)** — same in `_visit_classical_assignment`: L1984 raises for any `StretchType` lvalue before this line is reached.

src/pyqasm/visitor.py
```diff
@@ -193,6 +209,7 @@ def _construct_visit_map(self):
             qasm3_ast.CalibrationGrammarDeclaration: self._visit_calibration_grammar_declaration,
         }
 
+    @semantic_check_gate
```
ryanhill1: **N1** — `semantic_check_gate` reads as "quantum gate" in this file. Something like `skip_if_check_only` or `_check_only_returns_empty` says what it does without the collision.

**N2** — worth type-hinting `func: Callable[..., Any]` while you're here; #346 is adding hints across these modules and this would land unhinted.

src/pyqasm/visitor.py
```diff
@@ -1715,17 +1703,15 @@ def _visit_constant_declaration(
             statement.init_expression = PulseValidator.make_complex_binary_expression(init_value)
 
         if isinstance(statement.init_expression, qasm3_ast.FunctionCall):
-            statement.init_expression = (
-                self._handle_function_init_expression(statement.init_expression, init_value)
-                or statement.init_expression
-            )
-        self._handle_extern_function_cleanup(statements, statement)
+            function_name = statement.init_expression.name.name
+            if function_name in FUNCTION_MAP and isinstance(init_value, (float, int)):
```
ryanhill1, line 1838: **N3** — the inlining looks behaviour-preserving. The old `helper(...) or statement.init_expression` fallback was equivalent because `FloatLiteral` is a truthy dataclass, so nothing was relying on the `or`.

Only note: this is now duplicated three times (here, L1943, L2113). If it grows a fourth case, a small `_maybe_fold_function_call(expr, value)` helper would be worth it — but three identical two-liners is arguably clearer than the indirection you removed, so no objection.

src/pyqasm/visitor.py
```diff
@@ -731,22 +728,24 @@ def _resolve_unindexed_reset_qubit(self, statement: qasm3_ast.QuantumReset) -> b
...
+    @semantic_check_gate
     def _visit_reset(self, statement: qasm3_ast.QuantumReset) -> list[qasm3_ast.QuantumReset]:
         """Visit a reset statement element.
 
         Args:
             statement (qasm3_ast.QuantumReset): The reset statement to visit.
 
         Returns:
-            None
+            list[qasm3_ast.QuantumReset] - A list of unrolled resets.
```
ryanhill1: **N4** — the rest of the file uses Google style with a colon: `list[qasm3_ast.QuantumReset]: A list of unrolled resets.` This one uses a dash.

src/pyqasm/pulse/validator.py
```diff
@@ -114,10 +114,10 @@ def validate_duration_or_stretch_statements(
         Generic validation function for DurationType and StretchType declarations or assignments.
 
         Args:
-            statement: The AST statement node
-            statement_type: The expected AST node type
-            base_type: The declared type (DurationType or StretchType)
-            rvalue: The initializer or assigned value
+            statement: The AST statement node.
+            statement_type: The expected AST node type.
```
ryanhill1: **N5** — pre-existing, not yours, but you're editing this block: `statement_type` isn't a parameter. It's derived inside the function at L128 (`statement_type = type(statement)`). Worth deleting the line while you're here.

src/pyqasm/pulse/validator.py
```diff
@@ -114,10 +114,10 @@ def validate_duration_or_stretch_statements(
...
+            base_type: The declared type, function does nothing if not DurationType or StretchType
```
ryanhill1: **N6** — this is the one line in the hunk that didn't get a trailing period, and the other three did.

src/pyqasm/visitor.py
micpap25: My bad here; I was trying to match L1686, L1882, and L2067

src/pyqasm/pulse/validator.py
```diff
@@ -114,11 +114,10 @@ def validate_duration_or_stretch_statements(
         Generic validation function for DurationType and StretchType declarations or assignments.
 
         Args:
-            statement: The AST statement node
-            statement_type: The expected AST node type
-            base_type: The declared type (DurationType or StretchType)
-            rvalue: The initializer or assigned value
-            global_scope: Global symbol table.
+            statement (Statement): The AST statement node.
+            base_type (Any): The declared type, function does nothing if not DurationType or StretchType.
```
ryanhill1: **G1 — new `line-too-long`, fails CI.**

```
src/pyqasm/pulse/validator.py:118:0: C0301: Line too long (105/100)
```

Adding `(Any)` and the trailing period (N5/N6) pushed it over. Wrapping the clause onto a continuation line matches how the rest of the file handles long `Args:` entries:

```python
base_type (Any): The declared type; the function does nothing if it is
    neither DurationType nor StretchType.
```

This is the only remaining CI failure — `visitor.py` is clean, and `mypy`/`black`/`isort` all pass.

src/pyqasm/visitor.py
```diff
@@ -83,6 +84,22 @@
 logger = logging.getLogger(__name__)
 logger.propagate = False
 
+F = TypeVar("F", bound=Callable[..., Any])
+
+
+def check_only_return_empty(func: F) -> F:
```
ryanhill1, line 93: **Confirming N1/N2 as fixed.** `F -> F` correctly preserves the wrapped signature (the earlier `Callable[..., Any]` would have erased it), and `# type: ignore` on the return is the standard escape for this pattern — mypy is clean.

Worth noting for whoever extends this: the decorator is only safe on methods whose *every* return should be suppressed under `check_only`. That's what R1 turned on. If a fourth method gets decorated later, the check is "does it have an early return that isn't itself gated" — a one-line AST sweep catches it.

ryanhill1 (commented):
Re-checked at `453cee4`. **All 15 threads are fixed** — resolved. This is a good round; the R1 resolution in particular is exactly right.

**Verified, not eyeballed:**

- **B1–B4** — `visitor.py` is now pylint-clean (was 11 errors). Moving the decorator *below* the two `# pylint: disable` comments restored the block scoping, and dropping it from the nested `_evaluate_case` closure fixed the switch tests.
- **R1** — you took the conservative route: decorator off the three methods with reachable early returns, explicit `if self._check_only: return []` restored at the tail. I re-ran the differential probe that originally caught this:

  ```
  validate() on a defcal-measure pulse program -> statements in _unrolled_ast
  main:      4  (…, QuantumMeasurementStatement)
  this PR:   4  (…, QuantumMeasurementStatement)   # was 3
  ```

  Parity with `main` restored. I also re-swept the 11 still-decorated methods: the only early returns left are the three I'd classified as safe (`_visit_phase_operation`, `_visit_classical_declaration` both delegate to already-gated methods; `_visit_classical_assignment` returns `[]` regardless). No nested-function applications remain.
- **D1** — both `StretchType` branches reverted.
- **N1/N2** — renamed, and `F = TypeVar("F", bound=Callable[..., Any])` preserves the wrapped signatures.

**Suite: 648 passed, 3 skipped. mypy, black, isort all clean.** Net −25 lines.

### One new thing, and three carried over

**G1** — the docstring fix introduced a `line-too-long` in `validator.py`. That's the only thing failing CI now; inline below.

Carried over from my first pass, none of them blocking on their own:

- **No tests.** You flagged this yourself, and #188 asks for coverage specifically so that refactors like R1 get caught by CI rather than by review. The three `check_only` + openpulse paths are the obvious candidates — there's currently no test that would have failed when R1 was live.
- **No CHANGELOG entry.**
- **#346 still conflicts.** It modifies `_handle_function_init_expression`; this deletes it. GitHub says `MERGEABLE` only because neither has landed yet — whichever goes second needs a rebase.

### One for @TheGupta2012 to settle

The decorator is now `check_only_return_empty`, following my **N1**. I hadn't seen at the time that you'd already suggested `semantic_check_gate` in an earlier round — so that rename overrode your call, which wasn't my intent. Happy either way; worth one of you picking so it doesn't churn again.

micpap25:
@ryanhill1 Let's get #346 merged first and I can do the rebasing here.

micpap25:
@ryanhill1 Is there a test file that you think would be appropriate for the tests for `visitor.py`?

ryanhill1:
@micpap25 There's no `tests/test_visitor.py` and I'd rather not add one — the suite is organized by QASM feature, not source module, so "tests for `visitor.py`" always means "tests for the feature you changed."

For this PR I don't think you need any. The existing suite covers it: I ran coverage on `main` and all 11 `check_only` branches the decorator replaces are already exercised, and the `_handle_function_init_expression` inlining is equivalent (the old `helper(...) or original` only diverges on a falsy return, and `FloatLiteral` is always truthy). Full suite passes on your branch apart from the two `test_validate_*_invalid_file` CLI failures, which fail the same way on `main`.

The exception is the two `StretchType` additions in `_visit_classical_declaration` and `_visit_classical_assignment` — those are unreachable. Both paths are rejected earlier by the "Assignment to 'stretch' type variable ... is not allowed" guards, and `_visit_constant_declaration` already handled both types (covered by `const stretch s = -20ns;` in `tests/qasm3/declarations/test_classical.py`). So there's no test you *could* write for them, which is really the answer to your question — could you drop them and keep this a clean refactor? Validated non-const stretch would make a good follow-up: relax the guards first, then tests go next to `test_stretch_type_error`.

Minor: you've now got 11 decorated methods and about a dozen still using the explicit check. `_visit_function_call` can't use the decorator (`check_only and not self._in_extern_function`), but the rest could — your call whether to finish the sweep here.

micpap25:
@ryanhill1 Noticed a typo introduced by #346, pausing until #371 is merged.

micpap25:
@ryanhill1 Re: your last comment, the StretchType additions were removed some time ago and as discussed it isn't possible to change to `check_only_return_empty` for the other functions. So I think this is good to merge.

micpap25:
@ryanhill1 bumping this

src/pyqasm/visitor.py
```diff
@@ -2972,8 +2945,6 @@ def _evaluate_case(statements):
 
             self._scope_manager.pop_scope()
             self._scope_manager.restore_context()
-            if self._check_only:
-                return []
             return result
```
ryanhill1, line 2948: **B1 — blocking: this drops a guard `check_only` still needs.**

Fixing B2 correctly removed the decorator from `_evaluate_case`, but the original `if self._check_only: return []` went with it — so `_visit_switch_statement` now has no `check_only` handling at all.

It isn't redundant. `_visit_calibration_definition` returns `[statement]` under `check_only` (L3346, reached whenever the defcal body is non-empty — an empty body short-circuits at L3326), and `CalibrationDefinition` isn't in `SWITCH_BLACKLIST_STMTS`:

```qasm
defcalgrammar "openpulse";
qubit[2] q;
int i = 1;
switch (i) { case 1 { defcal x $0 { barrier $0; } } default { h q[0]; } }
```

```
main:         switch returns 0 statements,  unrolled_ast = 2
this branch:  switch returns 1 statement,   unrolled_ast = 3
```

`accept()` writes that into `_unrolled_ast.statements` even on the `validate()` path, so it escapes.

Restoring the two lines matches main exactly. Worth a comment on *why* — bare, they read as redundant, which is presumably how they came to be dropped — and worth a regression test: nothing in the 844 covers `check_only` with a switch, which is how this got past both CI and my first three probes at it.

src/pyqasm/visitor.py
```diff
@@ -20,13 +20,14 @@
 """
 
 import copy
+import functools
```
ryanhill1, line 23: **N1** — `import functools` here and `from functools import partial` at L28. Either add `wraps` to the existing `from` import or use `functools.partial` throughout.

ryanhill1 (commented):
Thanks for this, and for your patience on the turnaround.

One blocking item and one nit — all inline.

PR #349, opened 2026-08-05
docs(base): fix incorrect device_qubits kwarg in QasmModule.unroll() docstring

sathichaitanyasaidurga:
Related to #89, but does not close it -- #89 asks for the documented
   kwargs to become explicit named parameters on unroll(), which is a
   larger, separate change. This PR only corrects the existing docstring
   to be accurate.

QasmModule.unroll()'s docstring lists device_qubits as a supported
**kwargs entry, but unroll() forwards kwargs directly to
QasmVisitor.__init__(), which has no device_qubits parameter.
Calling module.unroll(device_qubits=5) raises:

    TypeError: QasmVisitor.__init__() got an unexpected keyword argument 'device_qubits'

Device qubit validation is actually inferred automatically inside the
visitor, so it isn't something a caller can set through unroll().

Change:
- Removed the incorrect device_qubits line from the unroll() docstring.
- Added a Notes entry clarifying that device-qubit validation is inferred
  automatically, not configurable via unroll() kwargs.

Testing:
- Reproduced the TypeError on main before this change to confirm the
  docstring was inaccurate.
- Docstring-only change; no functional code touched.
- black/isort pass; py_compile confirms no syntax errors.


src/pyqasm/modules/base.py
```diff
@@ -608,7 +607,10 @@ def unroll(self, **kwargs):
 
         Notes:
             This method resets the module's qubit and classical bit counts before unrolling,
-            and sets them to -1 if an error occurs during unrolling.
+            and sets them to -1 if an error occurs during unrolling. The device qubit count
+            used for validation is inferred automatically during unrolling and is not a
+            user-configurable ``unroll()`` keyword argument.
+
         """
```
ryanhill1, line 663: Both halves of this are inaccurate, and the thing a reader actually needs is missing.

**"inferred automatically during unrolling"** — only when `consolidate_qubits=True`. The inference at `visitor.py:3323-3328` is inside `if self._consolidate_qubits:`; a plain `unroll()` leaves `_device_qubits` as `None`.

**"not a user-configurable keyword argument"** — it is configurable, just on `loads()` rather than `unroll()` (`entrypoint.py:103-104`). And it matters:

```python
loads(src, device_qubits=5); unroll()                        # _device_qubits = 5
loads(src, device_qubits=5); unroll(consolidate_qubits=True) # ValidationError: Total qubits '(8)' exceed device qubits '(5)'
```

So this tells a reader the count is out of their hands, when the answer to "how do I set device qubits?" is "pass it to `loads()`" — which is why they're reading this docstring.

```suggestion
            This method resets the module's qubit and classical bit counts before unrolling,
            and sets them to -1 if an error occurs during unrolling. ``device_qubits`` is set
            on the module at load time -- ``loads(program, device_qubits=N)`` -- not through
            ``unroll()``. When unset and ``consolidate_qubits=True``, it defaults to the
            program's total qubit count.
        """
```

(The suggestion also drops the stray blank line before the closing quotes.)

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Types of changes:
 - `Dependencies`: for updates to external libraries or packages.
 
 ## Unreleased
+- Fixed inaccurate `device_qubits` entry in `QasmModule.unroll()` docstring ([#349](https://github.com/qBraid/pyqasm/pull/349))
```
ryanhill1: This lands directly under `## Unreleased`, above `### Added`, so it isn't in any category. Move it under `### Fixed` with the other entries.

ryanhill1 (commented):
Thanks for this — the core fix is right. Confirmed `QasmVisitor.__init__` has no `device_qubits` parameter and `unroll(device_qubits=5)` raises `TypeError`, and the five remaining documented kwargs all work, so the list is accurate once that line is gone.

Two things before merge:

1. The replacement Notes text needs a correction — details inline.
2. #89 asks for the kwargs to be promoted to explicit named parameters on `unroll()`, not just documented, so this doesn't close it as written. Either extend the PR to do that — very welcome, and the corrected docstring makes it straightforward — or drop the `Closes #89` line from the description.

CHANGELOG.md
sathichaitanyasaidurga: Thanks for the thorough review -- you're right on both points. Applied
your suggested docstring text, moved the changelog entry under Fixed,
and updated the PR description to no longer claim it closes #89 (it's
related but doesn't fully address the "promote to named params" ask).

ryanhill1 (approved):

PR #355, opened 2026-08-05
Rename duplicate argument variables internally

micpap25:
## Summary of changes

Fixes #311

Renames the variables created internally to handle quantum arguments in functions, preventing variable lookup issues caused by redefinition / shadowing.


micpap25:
@ryanhill1 I definitely don't think this is the most elegant way to handle this but the variable's name must be changed *before* the call to `transform_function_qubits` since the first thing that function does is use `_get_op_bits` which will break if it is passed the original register's name. Let me know if you think this is a good approach or if something better could be done.

src/pyqasm/subroutines.py
```diff
@@ -521,6 +521,11 @@ def process_quantum_arg(  # pylint: disable=too-many-locals
         """
         actual_arg_name = Qasm3SubroutineProcessor.get_fn_actual_arg_name(actual_arg)
         formal_reg_name = formal_arg.name.name
+        # If our actual variable is the same as the function argument,
+        # give the function argument a temporary name for internal use
+        if actual_arg_name == formal_reg_name:
+            formal_reg_name += "_arg"
```
ryanhill1: **This name reaches the user.** It flows into error messages, so a size mismatch reports `Expected 3 qubits in variable 'q_arg'` when the user wrote `q`. Breaks the 5 `test_subroutines.py` tests.

**Suggestion:** use a separate `internal_reg_name` for the map keys and leave `formal_reg_name` alone for messages.

src/pyqasm/subroutines.py
```diff
@@ -521,6 +521,11 @@ def process_quantum_arg(  # pylint: disable=too-many-locals
         """
         actual_arg_name = Qasm3SubroutineProcessor.get_fn_actual_arg_name(actual_arg)
         formal_reg_name = formal_arg.name.name
+        # If our actual variable is the same as the function argument,
+        # give the function argument a temporary name for internal use
+        if actual_arg_name == formal_reg_name:
```
ryanhill1, line 527: **`_arg` can collide with a name already in use:**

```qasm
qubit[2] q;
qubit[2] r;
def f(qubit q, qubit q_arg) { h q; x q_arg; }
f(q[0], r[1]);
```

`main` emits `h q[0]; x r[1];`. Here it's `ValueError: Variable 'q_arg' already exists in current scope` — and a raw `ValueError`, not a `ValidationError`.

**Suggestion:** use a name that can't appear in user code. `INTERNAL_QUBIT_REGISTER` in `elements.py` (`"__PYQASM_QUBITS__"`) is the existing pattern for this.

src/pyqasm/visitor.py
```diff
@@ -1499,11 +1499,21 @@ def _visit_generic_gate_operation(  # pylint: disable=too-many-branches, too-man
...
-                        operation, transform_map, size_map
+                try:
+                    operation.qubits = (
+                        Qasm3Transformer.transform_function_qubits(  # type: ignore [assignment]
+                            operation, transform_map, size_map
+                        )
+                    )
+                except KeyError:
+                    for qubit in operation.qubits:
+                        assert isinstance(qubit.name, str)
```
ryanhill1: **This fires on real programs** — it breaks `tests/cli/resources/valid1.qasm` and with it all 5 CLI tests:

```qasm
def create_bell_state(qubit[2] q) { hgate q[0]; cxgate q[0], q[1]; }
create_bell_state(q[0:2]);
```

A qubit operand has two shapes: for `Identifier`, `.name` is a `str`; for `IndexedIdentifier` (`q[0]`), `.name` is an `Identifier` and `.name.name` is the string. The assert only holds for the first.

Also worth knowing: assertions compile out under `python -O`, where this becomes `TypeError: unsupported operand type(s) for +=: 'Identifier' and 'str'`. Control flow is better as a real branch handling both shapes.

src/pyqasm/visitor.py
```diff
@@ -1499,11 +1499,21 @@ def _visit_generic_gate_operation(  # pylint: disable=too-many-branches, too-man
...
-                operation.qubits = (
-                    Qasm3Transformer.transform_function_qubits(  # type: ignore [assignment]
-                        operation, transform_map, size_map
+                try:
+                    operation.qubits = (
+                        Qasm3Transformer.transform_function_qubits(  # type: ignore [assignment]
+                            operation, transform_map, size_map
+                        )
+                    )
+                except KeyError:
```
ryanhill1, line 1635: **This is compensated here, but needed in three places.** The rename happens in `subroutines.py`, so `qubit_transform_map` is keyed on `q_arg` while the function body still says `q`. This `try/except` patches the mismatch at one use site — but there are three:

```
visitor.py:754    _visit_reset                    <- not patched
visitor.py:871    _visit_barrier                  <- not patched
visitor.py:1504   _visit_generic_gate_operation
```

So these still fail, now with a raw `KeyError` instead of the located `ValidationError` they gave before:

```qasm
def f(qubit q) { reset q; }                      // KeyError: ('q', 0)   (barrier too)

def inner(qubit p) { h p; }
def outer(qubit q) { inner(q); }
outer(q[1]);                                     // KeyError: 'q'  — works on main
```

**Suggestion:** keep the keys and the body consistent from the start — register the transform map under the name the body uses, or rename the body's references when you rename the parameter. Then no call site needs a retry and all three fix together.

Worth noting too: `except KeyError` catches it from anywhere inside `transform_function_qubits`, including genuine bugs.

ryanhill1 (commented):
Thanks for taking this on, @micpap25 — and for the root-cause work on #311. Identifying the parameter shadowing was the hard part, and the reported case does now work.

Most of what follows is in the `_visit_function_call` machinery you mentioned being unfamiliar with. Specifics are inline; two general things:

**The suite is red.** 10 tests fail here that pass on `main` — 5 in `tests/cli`, 5 in `tests/qasm3/subroutines`. `pytest tests` reproduces most of the inline findings.

**No tests or CHANGELOG entry.** A few regression tests would have caught most of this. Worth covering: the #311 repro, an array-typed shadowed parameter (`qubit[2] q`), `reset`/`barrier` bodies, a nested call with differing parameter names, and the `q`/`q_arg` collision. `tests/qasm3/subroutines/test_subroutines.py` is the natural home.

Your diagnosis was right — the gap is between fixing this program and fixing it without disturbing the others, and the four call sites in the last inline comment aren't discoverable without going looking. Happy to pair or split it, and no rush.

micpap25:
@ryanhill1 I'm going to start by building tests so that we do not revert back to failing the original #311 case.

micpap25:
Code has been updated so it's not failing any of the (existing) tests except for one which cares about the specific error (`ValueError`) being raised.

ryanhill1:
@micpap25 Great, thanks! Let me know when you're ready for a re-review

src/pyqasm/visitor.py
micpap25: Can you elaborate a bit on the approach you are describing here? I can implement the fix in three places, but I want to understand your suggestion. I'm a bit confused what you mean by "the body" for example.

micpap25:
@ryanhill1 Which file(s) should the tests be added to here?

ryanhill1:
@micpap25 `tests/qasm3/subroutines/test_subroutines.py` — and there's already a test for this exact bug sitting skipped in it:

```python
# line 347
@pytest.mark.skip(reason="Bug: qubit in function scope conflicts with global scope")
def test_qubit_renaming_in_formal_params():
```

I removed the skip and ran it: fails on `main`, passes on your branch. Deleting that line is the headline test here. I'd also add the minimal #311 repro (`def my_gate(qubit q) { h q; }` called as `my_gate(q[1])`) and a nested case next to `test_function_call_from_within_fn` at line 310.

First though, there's a regression to fix:

```
FAILED test_incorrect_custom_ops[redefinition_raises_error_2]
Failed: DID NOT RAISE ValidationError
```

That's `def my_function(qubit q) { int[32] q = 1; }`, which should report `"Re-declaration of variable 'q'"`. Since `process_quantum_arg` now registers the formal as `q__(QUANTUM_ARGUMENT)__`, the classical `q` no longer collides and the error disappears. I know you flagged this one, but I don't think it's the test being picky about the exception type — it's a real behaviour loss. Shadowing detection needs to keep using the user-facing name while lookup uses the internal one. Once it raises again, worth adding the mirror case to `SUBROUTINE_INCORRECT_TESTS` (a `qubit q` formal shadowed by a quantum re-declaration).

PR #357, opened 2026-08-07
fix: emit distinct x gates with fresh operands in negctrl expansion

### LINKED ISSUE #350: negctrl expansion emits the same QuantumGate object twice, aliasing operand nodes
## Description

The `negctrl` modifier expansion in `visitor.py` places the *same* `QuantumGate` object at two positions in the emitted statement list, rather than two equal copies:

```python
negs = [
    qasm3_ast.QuantumGate([], qasm3_ast.Identifier("x"), [], [ctrl]) for ctrl in negctrls
]
result = negs + result + negs
```

Every transformation that rewrites qubit indices in place (`remove_idle_qubits()`, `reverse_qubit_order()`) therefore reaches the same operand node more than once.

## Reproduction

```python
from pyqasm import loads

qasm = """OPENQASM 3.0;
include "stdgates.inc";
qubit[3] q;
negctrl @ x q[0], q[1];
"""
m = loads(qasm)
m.unroll()
print([id(s) for s in m._unrolled_ast.statements if hasattr(s, "qubits")])
# [4362757584, 4362758224, 4362757584]  <- statements 0 and 2 are the same object
m.reverse_qubit_order()  # KeyError: -2
```

`remove_idle_qubits()` on the same circuit survives only because of the `visited_node_ids` guard in `Modules._remap_qubits`; removing that guard raises `KeyError: 0`. Nothing in the test suite covers either path.

## Scope

Pre-existing, not a regression. #335 fixed the equivalent aliasing for decomposition-emitted statements (`maps/gates.py` constructors and `Decomposer`), but deliberately did not touch the modifier expansion path.

## Suggested fix

Build a second `negs` list for the trailing position rather than reusing the first, and run the control operand through `fresh_qubits()` so the two `x` gates own distinct operand nodes. Add a regression test asserting no shared statement or operand nodes after `unroll()` for `negctrl @ x`, and that `reverse_qubit_order()` succeeds on it.

Found in review of #335 (M2).

ryanhill1:
Fixes #350

The `negctrl @` expansion placed the *same* `QuantumGate` object at both the leading and trailing position of the emitted statement list. In-place transformations therefore reached the same operand nodes twice: `reverse_qubit_order()` raised `KeyError` on any unrolled `negctrl` gate, and `remove_idle_qubits()` survived only via the `visited_node_ids` guard.

The two `x` gates are now built as distinct statements whose operands go through `fresh_qubits()`, matching the fix #335 made for decomposition-emitted statements.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed negative-control gate expansion to generate correct leading and trailing `x` operations.
  * Resolved failures when reversing qubit order after unrolling negative-control gates.

* **Tests**
  * Added coverage for single- and multi-control negative-control operations.
  * Added regression testing for expected gate sequences and qubit remapping.

* **Documentation**
  * Added a changelog entry describing the fix.


ryanhill1:
@coderabbitai review

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Types of changes:
 ### Removed
 
 ### Fixed
+- Fixed the `negctrl @` expansion emitting the same `x` `QuantumGate` object at both the leading and trailing position, so in-place transformations mutated its operands twice — crashing `reverse_qubit_order()` (`KeyError`) on any unrolled `negctrl` gate. The two `x` gates are now distinct statements with fresh operand nodes. ([#350](https://github.com/qBraid/pyqasm/issues/350))
```
**Severity:** Medium

**Rationale:** This entry credits the fix with repairing `reverse_qubit_order()` only, but the same aliasing broke a second user-visible path that this PR also fixes. On `origin/main`, `unroll(consolidate_qubits=True)` raises `KeyError: '__PYQASM_QUBITS__'` for *any* `negctrl @` gate:

```
qubit[3] q;
negctrl @ x q[0], q[1];

# origin/main:  KeyError: '__PYQASM_QUBITS__'
#               (transformer.py:480, _get_pyqasm_device_qubit_index)
# 624ec7c4:     x __PYQASM_QUBITS__[0];
#               cx __PYQASM_QUBITS__[0], __PYQASM_QUBITS__[1];
#               x __PYQASM_QUBITS__[0];
```

Same root cause: `consolidate_qubit_registers` rewrote the shared `x` operand once, then met the already-renamed node a second time and looked `__PYQASM_QUBITS__` up in `global_qreg_size_map`. A reader scanning the changelog to see whether their crash is fixed will not find it under the current wording, so the release notes understate the change.

**Change Requested:** Name the second crash in the entry.

```suggestion
- Fixed the `negctrl @` expansion emitting the same `x` `QuantumGate` object at both the leading and trailing position, so in-place transformations mutated its operands twice — crashing `reverse_qubit_order()` (`KeyError`) and `unroll(consolidate_qubits=True)` (`KeyError: '__PYQASM_QUBITS__'`) on any unrolled `negctrl` gate. The two `x` gates are now distinct statements with fresh operand nodes. ([#350](https://github.com/qBraid/pyqasm/issues/350))
```

tests/qasm3/test_transformations.py
```diff
@@ -244,6 +246,31 @@ def test_unroll_emits_fresh_operand_nodes(operation):
     _assert_no_shared_operand_nodes(module)
 
 
+def test_reverse_qubit_order_negctrl():
```
TheGupta2012, line 254: **Type:** Implementation
**Severity:** Medium

**Rationale:** This test covers one of the three consumers the aliasing broke. Issue #350 names `remove_idle_qubits()` explicitly, noting it survives only because of the `visited_node_ids` guard in `Modules._remap_qubits` — an incidental guard rather than a deliberate contract, so nothing stops a later refactor from dropping it and silently reintroducing the crash. And `unroll(consolidate_qubits=True)` did not merely survive on `main`; it raised `KeyError: '__PYQASM_QUBITS__'` outright. Both paths are fixed by this change and neither is asserted, which leaves the fix resting on a single pass.

**Change Requested:** Add two short cases alongside this one, so all three passes the aliasing reached are pinned:

- `remove_idle_qubits()` on the unrolled `negctrl @ x q[0], q[1];` — expect `qubit[2] q; x q[0]; cx q[0], q[1]; x q[0];`
- `unroll(consolidate_qubits=True)` on the same program — expect `qubit[3] __PYQASM_QUBITS__; x __PYQASM_QUBITS__[0]; cx __PYQASM_QUBITS__[0], __PYQASM_QUBITS__[1]; x __PYQASM_QUBITS__[0];`

Both expected outputs were confirmed against this commit.

TheGupta2012 (commented):
## Verdict: approve with minor comments

The fix is correct, minimal, and consistent with the precedent #335 set for decomposition-emitted statements. It resolves issue #350 and, as a side effect, a second crash the description does not mention. Nothing here blocks merge. Four comments below: two Medium about documentation and coverage of that second crash, one Low about test breadth.

No security findings. The change adds no input handling, no I/O and no dependency, and `fresh_qubits()` was already public within `pyqasm.maps.gates`. No public API or shared contract moves, so the blast radius stays inside this repository.

### Findings

- [ ] **Medium** — the fix also repairs `unroll(consolidate_qubits=True)` crashing with `KeyError: '__PYQASM_QUBITS__'` on any `negctrl @` gate, which the CHANGELOG entry does not mention.
- [ ] **Medium** — regression coverage exercises `reverse_qubit_order()` only. `remove_idle_qubits()` is named in #350 and `consolidate_qubits=True` crashed outright on `main`; neither is asserted.
- [ ] **Low** — the `test_unroll_emits_fresh_operand_nodes` matrix gains two plain `negctrl` rows, while six other operand shapes the aliasing reached stay uncovered.

### How it was tested

Two worktrees were compared: PR head `624ec7c4` and `origin/main` at `7c31308`. Every claim below comes from running code.

**Issue #350 reproduces on `main` and is fixed here.** For one, two and three negative controls, `reverse_qubit_order()` after `unroll()` raises `KeyError: -5` on `main` and returns correct output on this branch.

**Object identity.** Every statement, operand node and index node in the unrolled AST was compared by `id()`. On `main`, the leading and trailing `x` are one object with one shared operand node and one shared index node per negative control. On this branch, zero shared objects across every case tried: one, two and three negative controls, `ctrl @ negctrl @`, `negctrl @ ctrl @`, `pow(2)` and `pow(3)` outer modifiers, a user-defined gate, physical qubits (`$0`), a broadcast register control (`negctrl @ x q, r;`), a discrete-set control (`q[{0,1}]`), `negctrl` inside a `gate` body invoked twice, inside a `for` body, inside an `if` body, and inside a `box` body.

**Semantics are unchanged.** The unrolled text this branch emits is byte-identical to `main` on all 14 cases `main` can unroll, so the `x … x` sandwich keeps its order. Independently, `qiskit.quantum_info.Operator` was compared for `negctrl @ <gate>` against its pyqasm expansion across 15 programs (qiskit 2.1.2, `qiskit-qasm3-import` 0.6.0). All 15 match exactly, not merely up to global phase. That set includes the `reverse_qubit_order()` output, checked against a hand-reversed reference program, and the `inv @ negctrl @` and `negctrl @ inv @` combinations on `rz` — `s` cannot serve here, since controlled `sdg` is unsupported on both worktrees for unrelated reasons.

**Suite and lint.** `pytest tests` on this branch: 717 passed, 2 failed, 4 skipped. On `main`: 714 passed, 2 failed, 4 skipped. The two failures are the same pre-existing `tests/cli/test_cli_commands.py` cases on both sides, so this PR adds no failure and adds 3 tests. `black --check` and `isort --check-only` are clean. `pylint` on `src/pyqasm/visitor.py` and on `tests/qasm3/test_transformations.py` produces no message this branch does not already produce on `main`; the changed region is clean. `mypy src` reports no error in `visitor.py`, and the `# type: ignore` on line 1621 was checked by removal and is still required. All 26 checks on `gh pr checks 357` pass. The `headers` lint target was not run, though the PR adds no file.

### Next steps

1. Extend the CHANGELOG entry to name the `consolidate_qubits=True` crash — a suggestion block is inline.
2. Add the two missing pass assertions to the new regression test.
3. Optionally widen the `test_unroll_emits_fresh_operand_nodes` matrix.

tests/qasm3/test_transformations.py
```diff
@@ -228,6 +228,8 @@ def _assert_no_shared_operand_nodes(module):
         "c4x q[0], q[1], q[2], q[3], q[4];",
         "ecr q[0], q[1];",
         "inv @ crz(0.5) q[1], q[2];",
+        "negctrl @ x q[0], q[1];",
+        "negctrl(2) @ x q[0], q[1], q[2];",
```
TheGupta2012, line 232: **Type:** Maintenance
**Severity:** Low

**Rationale:** These two rows cover the plain `negctrl` shape only. Testing on `origin/main` shows the aliasing also reached six further operand shapes that this change fixes but that no test pins: a broadcast register control (`negctrl @ x q, r;`), a discrete-set control (`negctrl @ x q[{0,1}], q[2];`), `negctrl` inside a `gate` body invoked twice, `negctrl` inside a `for` body, `pow(n) @ negctrl @`, and `ctrl @ negctrl @`. Each takes a different `_copy_qubit` branch or repeats the expansion, so each is a distinct way for the pattern to regress. This is the fourth fix for one bug class (#331, #333, #335, #350), which argues for making this matrix the durable guard rather than adding a bespoke test per incident.

**Change Requested:** Optional. Consider adding two or three of the above as parametrize rows — one line each, no new helper needed.

CHANGELOG.md
ryanhill1: Confirmed and applied in 32e1527. Reproduced on `origin/main`: `unroll(consolidate_qubits=True)` on `negctrl @ x q[0], q[1];` raises `KeyError: '__PYQASM_QUBITS__'`, and it returns the correct consolidated output on this branch. Took your wording verbatim.

tests/qasm3/test_transformations.py
ryanhill1: Both added in 32e1527, with your expected outputs confirmed locally.

`test_consolidate_qubits_negctrl` is a genuine regression test — it raises `KeyError: '__PYQASM_QUBITS__'` on main. `test_remove_idle_qubits_negctrl` passes on main too, exactly as you say, since `visited_node_ids` rescues it; the docstring records that so nobody later reads a green test as proof the guard is unnecessary.

One correction to the expected consolidated output: the register declaration is emitted *before* the include, so it is `qubit[3] __PYQASM_QUBITS__;` then `include "stdgates.inc";`.

tests/qasm3/test_transformations.py
ryanhill1: Agreed on making the matrix the durable guard rather than adding a bespoke test per incident. Added three rows in 32e1527:

```
negctrl @ x q[{0, 1}], q[2];
pow(2) @ negctrl @ x q[0], q[1];
ctrl @ negctrl @ x q[0], q[1], q[2];
```

I checked each against `origin/main` first — all three report a shared operand node there and are clean on this branch, so they are pulling their weight rather than just padding the matrix. (`inv @ negctrl @` also fails on main, but it takes the same path as `pow`, so I left it out.)

I skipped the broadcast-register and gate-body/for-body shapes: the parametrize fixture is a single operation against one `qubit[5] q`, and those need a second register or a multi-statement body, so they would need a separate test rather than a row. Happy to add them if you think the coverage gap is worth the extra test.

TheGupta2012 (approved):

PR #358, opened 2026-08-07
fix: drop unroll-emitted global phase for QASM 2 targets

### LINKED ISSUE #351: gphase emitted for QASM 2 modules: rzz/rxx unroll to a global phase QASM 2 cannot express
## Description

`rzz` and `rxx` decompose to a body containing a `QuantumPhase` node, which `Qasm2Module` emits as `gphase(...)`. OpenQASM 2 has no global-phase syntax at all, so the unroller turns a legal qelib1 gate into output no QASM 2 parser accepts.

## Reproduction

```python
from pyqasm import dumps, loads

src = """OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
creg m[1];
measure q[0] -> m[0];
if(m==1) rzz(0.3) q[0],q[1];
"""
m = loads(src)
m.unroll()
print(dumps(m))
```

```
if (m[0] == true) {
  gphase(-0.15) q[0], q[1];
  cx q[0], q[1];
  ...
}
```

The same applies outside a conditional — `rzz(0.3) q[0],q[1];` at top level unrolls to a `gphase` statement.

`QuantumPhase` is also absent from `Qasm2Module._whitelist_statements`, so a second filtering pass over an already-unrolled AST rejects it by AST class name. That is reachable today: `remove_idle_qubits()` and `reverse_qubit_order()` both reassign `_statements = _unrolled_ast.statements`, and `unroll()` has no re-entry guard, so `reverse_qubit_order()` followed by `remove_idle_qubits()` raises where it passes on `main`.

Sweeping ~45 qelib1/pyqasm gate shapes: every gate but `rzz`/`rxx` produces only `QuantumGate` nodes.

## Suggested fix

Either the unroller should not emit `gphase` for a `Qasm2Module` (global phase is unobservable, so dropping it is semantically safe for a QASM 2 target), or `_qasm_ast_to_str` should fold/drop the node at serialization. Whichever is chosen, add `QuantumPhase` handling to `_whitelist_statements` so the re-filtering path stops depending on it being absent.

#339 adds an interim `QuantumPhase`-specific diagnostic in `_filter_branch_body` so the error names global phase and points at `rzz`/`rxx`, rather than reporting an AST class for a program the user wrote as `rzz`. That is a message fix only — the underlying gap is this issue.

Found in review of #339 (M1).

ryanhill1:
Fixes #351

`rzz`/`rxx` decompose to a body containing a `QuantumPhase` node, which `Qasm2Module` emitted as `gphase(...)` — syntax OpenQASM 2 does not have, so unrolling a legal qelib1 gate produced output no QASM 2 parser accepts. The same node also broke re-filtering after `reverse_qubit_order()`/`remove_idle_qubits()`.

Global phase is unobservable, so `Qasm2Module.accept()` now strips `QuantumPhase` statements (descending into conditional bodies) from the unrolled AST. A user-written `gphase` in QASM 2 source is still rejected, and the #339 interim diagnostic is retained for that case with its message trimmed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed QASM 2 transformations involving conditional `rzz` and `rxx` operations.
  * Generated global-phase operations are now removed during unrolling, while user-authored `gphase` statements remain unsupported and are rejected.
  * Improved compatibility with idle-qubit removal and repeated reload/unroll workflows.

* **Tests**
  * Added regression coverage for operation decomposition, conditional transformations, and round-trip processing.


ryanhill1:
@coderabbitai review

ryanhill1:
@coderabbitai review

ryanhill1:
@coderabbitai review

ryanhill1:
Applied the changelog wording fix. Skipping the `-> None` / `Returns` annotations on test functions: no test in this repo's suite annotates `-> None` or documents a `Returns` section (0 occurrences across `tests/`), so adding it to only these six would be inconsistent with the established convention.

src/pyqasm/modules/qasm2.py
```diff
@@ -149,6 +148,21 @@ def to_qasm3(self, as_str: bool = False) -> str | Qasm3Module:
...
+        global-phase syntax, and a global phase is unobservable, so dropping it is
+        semantically safe (issue #351)."""
+        filtered = []
+        for stmt in statements:
+            if isinstance(stmt, qasm3_ast.QuantumPhase):
+                continue
+            if isinstance(stmt, qasm3_ast.BranchingStatement):
+                stmt.if_block = self._drop_global_phase(stmt.if_block)
+                stmt.else_block = self._drop_global_phase(stmt.else_block)
+            filtered.append(stmt)
```
TheGupta2012, line 187: **Type:** Implementation &nbsp;|&nbsp; **Severity:** Medium

**Rationale:** The drop can empty a conditional body, and nothing then removes the now-bodiless branch. A gate whose body is nothing but a phase reaches this:

```
gate ph(t) a { gphase(t); }
if(c==1) ph(0.3) q[1];
```

On this branch that unrolls to `if (c[0] == true) {` followed by a bare closing brace — an `if` with no `qop`, which QASM 2 has no form for. `remove_idle_qubits()` and `reverse_qubit_order()` both preserve it. `main` was wrong here too, but loudly: it emitted `gphase(0.3) q[1]` inside the branch and raised `ValidationError` on reload. This PR turns a rejection into silently malformed output, which is the worse failure mode.

Narrow reachability, granted — `rzz`/`rxx` never hit it, since their decompositions carry real gates alongside the phase. But #345 already established the precedent for exactly this shape: a `box` left empty by a removal is dropped, "since pyqasm rejects a box with no statements".

**Change requested:** Drop the branch when *both* blocks come back empty. Guarding on `if_block` alone would be wrong if an `else_block` ever survives.

```suggestion
            if isinstance(stmt, qasm3_ast.BranchingStatement):
                stmt.if_block = self._drop_global_phase(stmt.if_block)
                stmt.else_block = self._drop_global_phase(stmt.else_block)
                if not stmt.if_block and not stmt.else_block:
                    # the body was nothing but global phase, and QASM 2 has no
                    # form for a conditional without a qop
                    continue
            filtered.append(stmt)
```

src/pyqasm/modules/qasm2.py
```diff
@@ -149,6 +148,21 @@ def to_qasm3(self, as_str: bool = False) -> str | Qasm3Module:
...
 
+    def _drop_global_phase(self, statements):
+        """Remove QuantumPhase statements the unroller emitted (e.g. from the rzz/rxx
+        decompositions), descending into conditional bodies. OpenQASM 2 has no
+        global-phase syntax, and a global phase is unobservable, so dropping it is
+        semantically safe (issue #351)."""
+        filtered = []
+        for stmt in statements:
+            if isinstance(stmt, qasm3_ast.QuantumPhase):
+                continue
```
TheGupta2012, line 179: **Type:** Maintenance &nbsp;|&nbsp; **Severity:** Low

**Rationale:** The drop keys on the node type alone. It is correct today, and that was confirmed rather than assumed: `_visit_phase_operation` in `visitor.py` clears `.modifiers` and rewrites any *controlled* phase into a `p` gate, so nothing carrying a control modifier ever arrives here. That invariant lives in another module, though, and nothing at this site records the dependency. Should it ever weaken, a controlled phase — a **relative** phase, and fully observable — would vanish with no error and no visible signal. Silent physical incorrectness is expensive to find later.

**Change requested:** Assert the invariant instead of relying on it. `raise_qasm3_error` is already imported at line 26.

```suggestion
            if isinstance(stmt, qasm3_ast.QuantumPhase):
                # a controlled phase is relative, not global, and is observable;
                # the visitor rewrites those to 'p' gates, so none should reach here
                if stmt.modifiers:
                    raise_qasm3_error(
                        "Modified global phase cannot be dropped for a QASM 2 target",
                        error_node=stmt,
                        span=stmt.span,
                    )
                continue
```

tests/qasm2/test_conditional_body.py
```diff
@@ -64,15 +64,13 @@ def test_conditional_non_qop_rejected(operation, keyword):
...
-    unrolled body, which remove_idle_qubits/reverse_qubit_order do (issue #351)."""
+def test_conditional_rzz_survives_refiltering():
+    """Test that transformations which re-filter an already unrolled body no longer
+    trip over the rzz global phase: it is dropped for a QASM 2 target (issue #351)"""
     module = loads(QASM2_PREAMBLE + "if(m==1) rzz(0.3) q[0], q[1];\n")
     module.unroll()
     module.reverse_qubit_order()
-    with pytest.raises(ValidationError, match="Global phase is not representable in QASM 2.0"):
-        module.remove_idle_qubits()
+    module.remove_idle_qubits()
```
TheGupta2012, line 73: **Type:** Maintenance &nbsp;|&nbsp; **Severity:** Medium

**Rationale:** This PR edits the wording of the conditional-body `gphase` diagnostic (`qasm2.py:101-106`) and, in the same change, removes the only test that exercised it. `grep -rn "not representable in QASM 2.0" tests/` returns nothing on this branch; on `main` it matched the `pytest.raises(...)` this test replaced.

The new `test_user_written_gphase_rejected` in `test_operations.py` does not close the gap: a top-level `gphase` is caught by the whitelist in `_filter_statements` and produces a *different* message (`Statement of type <class 'openqasm3.ast.QuantumPhase'> not supported in QASM 2.0`). The conditional-body path is still live — `if(m==1) gphase(0.3);` was confirmed to raise the trimmed message under both `validate()` and `unroll()` — but it is now untested, immediately after its text was changed.

**Change requested:** Keep this test as written, and add one alongside it for the rejection path, so the branch this PR reworded stays pinned:

```python
def test_conditional_user_gphase_rejected():
    """A gphase the user wrote in a conditional body is still rejected (issue #351)."""
    module = loads(QASM2_PREAMBLE + "if(m==1) gphase(0.3);\n")
    with pytest.raises(ValidationError, match="Global phase is not representable in QASM 2.0"):
        module.validate()
```

tests/qasm2/test_operations.py
```diff
@@ -67,6 +67,113 @@ def test_whitelisted_ops():
...
+    if(m==1) rzz(0.3) q[0], q[1];
+    """
+
+    result = loads(qasm2_string)
+    result.unroll()
+    unrolled = dumps(result)
+    assert "gphase" not in unrolled
+
+    # the unrolled output must be a loadable QASM 2 program
+    loads(unrolled).validate()
```
TheGupta2012: **Type:** Implementation &nbsp;|&nbsp; **Severity:** Low

**Rationale:** The comment claims more than the assertion establishes. `loads(...).validate()` proves pyqasm re-reads its own output; it does not prove the output is QASM 2. It is not: `qiskit.qasm2.loads` rejects this exact string with `QASM2ParseError: needed '==', but instead saw [`, because the conditional prints as `if (c[0] == true) { ... }` — QASM 3 syntax. Emitting `gphase` was one reason this output would not load, but not the only one.

That remaining half is #338's territory rather than this PR's, so no code change is needed here. A future reader taking the comment at face value would conclude the conditional path is fully fixed, though, and it is not.

**Change requested:** Narrow the comment to what is actually asserted — for example `# the unrolled output must still re-load in pyqasm` — or keep the stronger claim and reference #338 as the remaining blocker.

TheGupta2012 (commented):
### Verdict

**Approve with non-blocking comments** (carried as prose here; this review is posted as a comment, not as a GitHub approval). The fix is correct and well-scoped. The central safety question — whether dropping `QuantumPhase` can ever discard an *observable* phase — was tested adversarially, and the answer is no, for a reason that holds structurally rather than by luck: `_visit_phase_operation` (`src/pyqasm/visitor.py:1430-1445`) rewrites a controlled phase into a `p` gate and clears `.modifiers`, so every `QuantumPhase` reaching `_drop_global_phase` is a genuine global phase. The two Medium findings below are worth addressing before merge; neither blocks.

### Findings

| # | Type | Severity | Summary |
|---|---|---|---|
| 1 | Implementation | Medium | A conditional whose body is entirely global phase is emitted as a bodiless `if`, silently, where `main` raised |
| 2 | Maintenance | Medium | The conditional-body `gphase` diagnostic had its message edited and its only test deleted |
| 3 | Maintenance | Low | The drop is `isinstance`-only; correctness rests on an invariant held in `visitor.py` and undocumented at the drop site |
| 4 | Implementation | Low | A test comment claims the output is "a loadable QASM 2 program"; an external QASM 2 parser rejects it |

### How this was verified

Every claim below comes from executing code in a worktree at `1246c656`, compared against a second worktree at `origin/main` (`7c31308`).

**Global phase is genuinely global — the load-bearing check.** Three independent lines of evidence:

- `ctrl @ rzz(0.3)` and `ctrl @ rxx(0.3)`, and their `negctrl @` forms, are rejected upstream — `ValidationError: Unsupported controlled QASM operation: rzz with 1 controls` — on both `main` and this branch. Wrapping `rzz` in a user-defined gate and controlling that wrapper is rejected identically.
- Controlling a gate that *does* carry a phase reaches `_visit_phase_operation` with a non-empty `ctrls`, which converts it into a `p` gate. For `gate myg(t) a { gphase(t); x a; }` applied under `ctrl @`, zero `QuantumPhase` nodes survive and the emitted body is `h; rx(0.3); h; cx` — the phase became a real rotation rather than a dropped node.
- Numerically, that emitted circuit matches `Operator(UnitaryGate(exp(i*theta)*X).control(1))` under `.equiv()`. The repository's own `tests/qasm3/test_gates.py:768 test_ctrl_gphase_eq_p` already pins the same `ctrl @ gphase == p` identity.

**Re-filter matrix**, for `rzz` and `rxx`, five sequences each: `remove_idle_qubits()`→`validate()`; `remove_idle_qubits()` twice; `remove_idle_qubits()`→`reverse_qubit_order()`; `reverse_qubit_order()`→`remove_idle_qubits()`; `unroll()`→`remove_measurements()`→`validate()`. All 10 fail on `main` with `Statement of type <class 'openqasm3.ast.QuantumPhase'> not supported in QASM 2.0`. All 10 pass here, and none of the outputs contain `gphase`.

**Emitted output under an external parser.** `qiskit.qasm2.loads(..., custom_instructions=LEGACY_CUSTOM_INSTRUCTIONS)` accepts the unrolled `rzz` and `rxx` programs; it rejects `main`'s output. Both decompositions do drop a nonzero phase — `gphase(-0.15)` for `rzz(0.3)` and `gphase(-0.35)` for `rxx(0.7)`, i.e. `-theta/2` in each case. `Operator(...).equiv()` confirms the emitted QASM 2 matches qiskit's `RZZGate(0.3)` / `RXXGate(0.7)` and matches pyqasm's own QASM 3 unrolling of the same source, up to global phase in every pairing — exactly the intended difference and nothing more.

**Rejection of user-written `gphase`** still holds in both positions: at top level via the whitelist in `_filter_statements`, and inside a conditional via the retained diagnostic. Both were confirmed under `validate()` and under `unroll()`.

**Containers.** `_drop_global_phase` descends into `if_block`/`else_block` only, and that is sufficient rather than merely untested: `_filter_statements` rejects both `Box` and `Pragma` for a QASM 2 module (`Statement of type <class 'openqasm3.ast.Box'> not supported in QASM 2.0`), so the `#pragma braket verbatim` box path added in #341 cannot reach a `Qasm2Module`. `BranchingStatement` is the only whitelisted container, and the unroller expands loops.

**No aliasing consequence** follows from mutating `stmt.if_block` in place: unrolling the same module object twice is byte-identical, and matches a freshly loaded module.

**Suite, lint, CI.** Full `pytest`: 719 passed here against 714 on `main`, with the same two pre-existing `tests/cli/test_cli_commands.py` failures on both, which are environmental and unrelated. `black --check` is clean. All 26 CI checks on the PR report success.

### Not verified

`isort` and `pylint` are not present in the available environment, so `black` was the only formatter run locally; CI covers the remainder.

### Next steps

- **Merge-order conflict with #338.** Open PR #338 ("fix: emit OpenQASM 2 syntax for classical conditionals") rewrites this same file with +364 lines, adding `_filter_branch`, `Qasm2Printer` and `_flatten_branch`. Whichever of the two merges second will conflict in `src/pyqasm/modules/qasm2.py`. Finding 1 interacts with it directly, since #338 introduces the printer that would have to render a branch emptied by the phase drop. Sequencing the two deliberately will cost less than resolving that conflict blind.
- Pre-existing and out of scope, but adjacent enough to deserve an issue: `inv @ rzz(0.3) q[0], q[1];` raises `AttributeError: 'QuantumPhase' object has no attribute 'name'` on both `main` and this branch.

src/pyqasm/modules/qasm2.py
ryanhill1: Confirmed and applied in 85620fe. Your repro produces exactly `if (m[0] == true) {` + `}` on this branch, and you are right that trading main's loud rejection for silently malformed output is the worse direction. Took your guard as written — both blocks empty, not `if_block` alone. Since `_drop_global_phase` recurses before the check, nested emptied branches collapse bottom-up too.

Pinned as `test_conditional_emptied_by_phase_drop_is_removed`, and noted in the changelog entry.

src/pyqasm/modules/qasm2.py
ryanhill1: Applied in 85620fe. Fair point that the invariant lives in `_visit_phase_operation` and nothing at this site records the dependency — a silently dropped relative phase is exactly the kind of thing that costs a day to find. Took the guard verbatim.

tests/qasm2/test_conditional_body.py
ryanhill1: Good catch — I reworded that diagnostic and deleted its only test in the same commit. Added `test_conditional_user_gphase_rejected` in 85620fe, essentially as you wrote it. Confirmed the path is live: `if(m==1) gphase(0.3);` raises the trimmed message under both `validate()` and `unroll()`.

tests/qasm2/test_operations.py
ryanhill1: You are right, and I checked rather than took it on faith: `qiskit.qasm2.loads` rejects that output with `QASM2ParseError: needed '==', but instead saw [`, because the conditional prints as `if (m[0] == true) { ... }`.

Narrowed the comment in 85620fe and named #338 as the remaining blocker, so the next reader does not mistake this for the conditional path being fully fixed.

TheGupta2012 (approved):

PR #359, opened 2026-08-07
fix: record an external gate's own depth, not its skipped decomposition's

### LINKED ISSUE #352: External/verbatim gates report the depth of the decomposition they skipped
## Description

`_visit_external_gate_operation` calls `_visit_basic_gate_operation(operation)` "just for validation" (`src/pyqasm/visitor.py:1351`), and that call still runs `_update_qubit_depth_for_gate` over the decomposition. So a gate that is *not* decomposed because it is external still contributes the depth of the decomposition that was skipped.

The `_recording_ext_gate_depth` suppression on the neighbouring line guards only the custom-gate path, not the basic-gate path.

## Reproduction

```python
from pyqasm import loads

qasm = """OPENQASM 3.0;
include "stdgates.inc";
qubit[2] q;
crz(0.5) q[0], q[1];
"""
m = loads(qasm)
m.unroll(external_gates=["crz"])
print(m.depth())   # 12, but exactly one crz statement is emitted
```

Only gates with a decomposition rule are affected — `unroll(external_gates=["ccx"])` reports 1 correctly, as does a body of `rx` + `cz` (2).

## Why it matters more now

Before #341 this required opting in per gate name via `external_gates`. #341 routes every gate inside a `#pragma braket verbatim` box through the same path, so any Braket verbatim program containing a decomposable gate silently reports an inflated depth without the user naming anything:

```
#pragma braket verbatim
box { crz(0.5) q[0], q[1]; }
```

emits one instruction and reports `depth == 12`.

## Suggested fix

Suppress depth recording around the validation-only `_visit_basic_gate_operation` call, the same way `_recording_ext_gate_depth` already does for the custom-gate path, and record the external gate's own depth instead.

Found in review of #341 (M1).

ryanhill1:
Fixes #352

`_visit_external_gate_operation` calls `_visit_basic_gate_operation` for validation only, but that call still ran `_update_qubit_depth_for_gate` over the decomposition that was never emitted — so `unroll(external_gates=["crz"])` on a single `crz` reported `depth() == 12` while emitting one statement. Since #341 the same path handles every decomposable gate inside a `#pragma braket verbatim` box.

Depth recording is now suppressed around the validation-only call (the same `_recording_ext_gate_depth` mechanism the custom-gate path uses), and the external gate records its own depth — including the branch-qubit marking variant inside conditionals.


ryanhill1:
@coderabbitai review

ryanhill1:
@coderabbitai review

ryanhill1:
@argus-eye review

src/pyqasm/visitor.py
```diff
@@ -1399,6 +1406,16 @@ def gate_function(*qubits):
         all_targets = self._unroll_multiple_target_qubits(operation, gate_qubit_count)
         result = self._broadcast_gate_operation(gate_function, all_targets)
 
+        # record the external gate's own depth; the custom-gate path has already done so
+        if gate_name not in self._custom_gates:
+            if not self._in_branching_statement:
+                self._update_qubit_depth_for_gate(all_targets, ctrls)
+            else:
```
TheGupta2012, line 1438: **Type:** Implementation
**Severity:** Low

**Rationale:** This branching arm is reachable but has no observable effect today, and no test covers it.

Evidence: replacing the whole `else` body with `pass` leaves the full PR suite green — 718 passed, 4 skipped, and only the two pre-existing `test_cli_commands.py` failures that also fail on base. Depth was also unchanged across six conditional programs (external gate in `if`, `ctrl @` external in `if`, external plus a second gate in `if`, external on qubits idle outside the branch, external gate inside an external custom gate in `if`, and a nested `if`) — every one reports `2` on base, on this PR, and with the arm deleted.

The reason is that the arm duplicates marking the validation-only call already performs. `_recording_ext_gate_depth` guards only `_update_qubit_depth_for_gate` (visitor.py:1098); it does not guard `_mark_branch_qubit`. So the suppressed `_visit_basic_gate_operation` call still reaches the branch-marking block at visitor.py:1201-1208 and marks the same qubits.

This is worth keeping rather than deleting. The arm becomes load-bearing the moment the suppression is extended to cover branch marking too — which is the natural reading of "suppress depth recording" and a plausible future refactor. Right now nothing pins that behaviour, so a later change could silently regress conditional depth with no test failing.

**Change Requested:** Add a test that fixes the conditional contract, so the arm is covered and a future change to the guard cannot regress it silently. For example, in `tests/qasm3/test_depth.py`, assert that a `crz` inside an `if` block reports the same depth whether or not `crz` is external (both are `2` today), and ideally one case where the external gate acts on qubits untouched outside the branch.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Types of changes:
 ### Removed
 
 ### Fixed
+- Fixed external and verbatim-box gates counting the depth of the decomposition they skipped: `unroll(external_gates=["crz"])` on a single `crz` reported `depth() == 12` while emitting one statement. An external gate now records its own depth, like the custom-gate path already did. ([#352](https://github.com/qBraid/pyqasm/issues/352))
```
**Severity:** Low

**Rationale:** "like the custom-gate path already did" reads as though the custom-gate path is correct in general. It is not, for nested external custom gates — the same user-visible symptom this entry describes still occurs there.

Evidence, identical on base and on this PR (so **pre-existing and not a regression from this change**):

```qasm
gate inner a, b { crz(0.5) a, b; }
gate outer a, b { inner a, b; h a; }
qubit[2] q;
outer q[0], q[1];
```

With `external_gates=["outer"]` this emits exactly one statement, `outer q[0], q[1];`, but reports `depth() == 13`. With `external_gates=["inner", "outer"]` it reports `2` for the same single statement. A single-level external custom gate correctly reports `1`, which is why the existing tests pass.

The cause is precisely the pattern this PR improved on. `_visit_custom_gate_operation` assigns the flag unconditionally at visitor.py:1285 and clears it unconditionally at visitor.py:1317-1318, with no save-restore. The inner custom gate therefore clobbers the outer gate's flag, and the outer gate never records its own depth. The new code here avoids exactly that by saving `prev_recording` and restoring it in a `finally` — applying the same pattern at :1285 would close the remaining gap.

**Change Requested:** Soften the wording so the entry does not claim the custom-gate path is already fully correct. Consider a follow-up issue for the nested case; fixing it here would widen the PR beyond #352.

```suggestion
- Fixed external and verbatim-box gates counting the depth of the decomposition they skipped: `unroll(external_gates=["crz"])` on a single `crz` reported `depth() == 12` while emitting one statement. An external gate now records its own depth, matching how a single-level external custom gate is already handled. ([#352](https://github.com/qBraid/pyqasm/issues/352))
```

TheGupta2012 (dismissed):
## Verdict

**Approve.** The fix is correct, minimal, and well targeted. Suppressing depth recording around the validation-only `_visit_basic_gate_operation` call is the right mechanism, and the `try`/`finally` save-restore is more robust than the unconditional assignment the custom-gate path uses. Two Low-severity suggestions below; neither blocks merge.

The fix reaches further than the issue title suggests. Verified corrections beyond the single-`crz` repro:

| Program (`crz` external / verbatim) | base | this PR | statements emitted |
|---|---|---|---|
| single `crz` | 12 | **1** | 1 |
| two sequential `crz`, same qubits | 24 | **2** | 2 |
| two `crz`, disjoint qubits | 12 | **1** | 2 |
| `crz` in a `for` loop, 3 iterations | 36 | **3** | 3 |
| `crz` in a plain (non-verbatim) `box` | 12 | **1** | 1 |
| `crz` in a `#pragma braket verbatim` box | 12 | **1** | 1 |
| `crz` inside a plain custom gate body | 12 | **1** | 1 |

## Findings

- [ ] **Low / Implementation** — the new branching arm is untested and currently has no observable effect (inline on `src/pyqasm/visitor.py:1413`)
- [ ] **Low / Maintenance** — CHANGELOG wording overstates the custom-gate path's correctness (inline on `CHANGELOG.md:29`)

No Critical, High, or Medium findings. No security surface: the change touches only an internal depth metric, adds no I/O, and introduces no new parsing or user-controlled path.

## How this was tested

A detached worktree at `89453bcc` was compared against `origin/main` (`7c31308`) by running both trees through the same battery.

1. **Issue #352 repro** — confirmed base reports `depth() == 12` for one emitted statement; the PR reports `1`. The value `1` was derived independently from the emitted program, not read off the PR.
2. **Double-counting** — `_broadcast_gate_operation` records no depth, and no code after the new block does either. Two sequential external gates on the same qubits give `2`; on disjoint qubits, `1`; sharing one qubit, `2`. Counted exactly once.
3. **Broadcast form** — `crz(0.5) c, t;` over registers passes a multi-subset `all_targets`, a shape the custom-gate path never exercises. Depth is correct for each emitted chunk.
4. **Flag safety** — instrumented `_visit_external_gate_operation` to record `_recording_ext_gate_depth` on entry and exit across nested external gates, external gates inside plain and external custom gate bodies, three invalid external gate calls that raise `ValidationError`, and an injected `RuntimeError` inside the validation call. Zero leaks in every case; the `finally` holds.
5. **Regression sweep** — a 10-program corpus with no external gates (Bell, GHZ, decomposable gates, custom gates, measurement, physical qubits, barrier, reset, broadcast, `pow @`) reports **identical** depth on base and PR, for both `decompose_native_gates=True` and `False`. `depth(decompose_native_gates=False)` is unchanged across every case in the battery.
6. **Test suite** — 718 passed on the PR versus 714 on base (the 4 new tests), same 4 skips. Two `tests/cli/test_cli_commands.py` failures reproduce identically on base and are unrelated to this change.
7. **Lint and CI** — `black --check` clean on all three changed files; all 26 GitHub checks pass.

## Next steps

1. Add a conditional-path test, as described inline on `visitor.py:1413`.
2. Apply the one-line CHANGELOG suggestion.
3. Optionally open a follow-up issue for the nested external custom gate gap described inline on `CHANGELOG.md:29` — pre-existing, out of scope here.

src/pyqasm/visitor.py
ryanhill1: Added in ecd3cf5 — two parametrized cases in `test_depth.py`, one with the conditional gate on qubits also used outside the branch and one where it is the only thing touching them. Both assert depth 2 external and non-external.

Worth being precise about what they buy, though: I reproduced your `pass` experiment and the new tests **also** pass with the arm neutered, for exactly the reason you gave — the validation-only call still reaches `_mark_branch_qubit`. So these pin the conditional *contract*, not the arm's current effect. That is what you asked for and it is the useful half: if someone extends `_recording_ext_gate_depth` to cover branch marking, the arm becomes load-bearing and these tests start failing if it is wrong. But nothing here would catch the arm being deleted today, and I would rather that be on the record than imply more coverage than exists.

CHANGELOG.md
ryanhill1: Confirmed your repro exactly: `external_gates=["outer"]` reports `depth() == 13`, `["inner", "outer"]` reports `2`, and a single-level external custom gate reports `1` — identical on base and on this branch, so pre-existing rather than a regression here.

Took your wording in ecd3cf5. Agreed the fix belongs in a separate PR: `_visit_custom_gate_operation` needs the same save-restore around the flag at :1285/:1317 that this PR uses. Happy to open a follow-up issue for it unless you would rather file it with your repro.

TheGupta2012 (approved):

PR #361, opened 2026-08-07
fix: use presence tests for loads() kwargs and reject invalid ones

### LINKED ISSUE #356: loads() silently drops falsy kwarg values (device_qubits=0, device_cycle_time=0.0, ...)
## Description

`loads()` stores six of its documented kwargs with a walrus **truthiness** test rather than a presence test:

```python
if dev_qbts := kwargs.get("device_qubits"):
    module._device_qubits = dev_qbts
```

So any falsy value is silently treated as "not passed" — no error, no warning, the caller's argument is discarded. `entrypoint.py:103-116`.

The same function already uses the correct pattern for its two bool kwargs:

```python
if "frame_in_def_cal" in kwargs:
    module._frame_in_def_cal = kwargs["frame_in_def_cal"]
```

## Reproduction

```python
from pyqasm import loads

src = 'OPENQASM 3.0;\ninclude "stdgates.inc";\nqubit[2] q;\nh q[0];\n'

loads(src, device_qubits=0)._device_qubits                 # None  <- expected 0
loads(src, device_cycle_time=0.0)._device_cycle_time       # None  <- expected 0.0
loads(src, compiler_angle_type_size=0)._compiler_angle_type_size  # None
loads(src, frame_limit_per_port=0)._frame_limit_per_port   # None

loads(src, frame_in_def_cal=False)._frame_in_def_cal       # False <- correctly honoured
```

## Scope

Pre-existing on `main`; unrelated to any open PR.

`extern_functions={}` is affected by the same test but is harmless in practice — the attribute already defaults to `{}`, so the dropped value is indistinguishable from the default. The four `Optional[...] = None` attributes above are the ones where a caller's `0` becomes `None`.

Two adjacent gaps in the same block, worth deciding on together:

- **No value validation.** Negative values *are* truthy, so they are stored and surface later as a confusing message rather than being rejected at the call site:

  ```python
  m = loads(src, device_qubits=-5)
  m.validate()   # ValidationError: Total qubits '2' exceed device qubits '-5'.
  ```

- **Unknown kwargs are silently ignored.** `loads(src, devise_qubits=5)` raises nothing and produces a module with `_device_qubits = None`. A typo in a documented kwarg is invisible until the resulting behaviour is wrong.

## Suggested fix

Use presence tests consistently for all of them:

```python
if "device_qubits" in kwargs:
    module._device_qubits = kwargs["device_qubits"]
```

Optionally reject non-positive `device_qubits` / `device_cycle_time` / `frame_limit_per_port` at load time, and raise on unrecognised kwarg names, so mistakes fail where they are made.

Found while verifying #349, which documents `device_qubits` as a `loads()` kwarg.

ryanhill1:
Fixes #356

Six documented `loads()` kwargs were stored behind a walrus truthiness test (`if dev_qbts := kwargs.get("device_qubits")`), so any falsy caller value was silently treated as "not passed". The kwarg→attribute map (`_LOADS_KWARG_ATTRS`) replaces the seven hand-written `if` blocks, so the kwarg list, the docstring and the storage loop can no longer drift apart.

Storage now tests `is not None`. An explicit `None` means "not passed", so module defaults are never clobbered — pass `False` to turn off a boolean kwarg.

Both adjacent gaps from the issue are closed as well:

- unknown kwarg names (`loads(src, devise_qubits=5)`) raise `TypeError`. **This is breaking** for any caller currently passing a stray or misspelled kwarg, and is filed under `Improved / Modified` accordingly.
- `device_qubits` / `device_cycle_time` / `compiler_angle_type_size` / `frame_limit_per_port` are validated at load time instead of surfacing later as a confusing validation message: non-positive raises `ValueError`, non-numeric raises `TypeError`. `bool` is rejected explicitly, since it is an `int` subclass and `device_qubits=True` would otherwise store as a count of 1.

`load()` forwards `**kwargs`, so it validates too — its errors name `load()` rather than `loads()`, and its docstring documents the exceptions.

### Behaviour change worth calling out

`frame_in_def_cal=None` and `play_in_cal_block=None` previously **were** stored, and both are consumed as falsy guards, so an explicit `None` disabled the check that the `True` default enables. They now mean "not passed" and the default survives. Nothing is lost — `False` reaches the state `None` used to — but callers building kwargs from a config mapping should know. Covered in the CHANGELOG and the `loads()` docstring.


ryanhill1:
@argus-eye review

src/pyqasm/entrypoint.py
```diff
@@ -99,21 +133,12 @@ def loads(program: openqasm3.ast.Program | str, **kwargs) -> QasmModule:
...
-        module._frame_in_def_cal = kwargs["frame_in_def_cal"]
-    if frame_limit_per_port := kwargs.get("frame_limit_per_port"):
-        module._frame_limit_per_port = frame_limit_per_port
-    if "play_in_cal_block" in kwargs:
-        module._play_in_cal = kwargs["play_in_cal_block"]
+    # presence tests, not truthiness: a falsy value is a caller value, not an omission.
+    # An explicit None still means "not passed", so defaults like extern_functions={}
+    # are never clobbered.
+    for name, attr in _LOADS_KWARG_ATTRS.items():
+        if kwargs.get(name) is not None:
```
TheGupta2012, line 170: **Type:** Implementation
**Severity:** Medium

**Rationale:** For the two bool kwargs this is a silent behaviour reversal, not a no-op. On `main` the storage test was `if "frame_in_def_cal" in kwargs`, so an explicit `None` **was** stored. Under `is not None` it is discarded and the default `True` survives (`modules/base.py:150`, `:152`). Both attributes are consumed as falsy guards (`pulse/visitor.py:714`, `:728`), so the effective flag flips from off to on.

Verified end to end on a `defcal` containing `newframe`:

```
                                    main            this PR
frame_in_def_cal omitted            unroll OK       unroll OK
frame_in_def_cal=None               ValidationError unroll OK   <-- flipped
frame_in_def_cal=False              ValidationError ValidationError
```

A caller building kwargs from a config mapping (`cfg.get("frame_in_def_cal")`) silently gets the opposite policy, with no error and no warning. Nothing is lost in capability — `False` reaches the same state `None` used to — but the change is undocumented, and the PR description's auto-generated section asserts the opposite ("Stores explicitly provided `None` values for supported `loads()` kwargs"). `tests/test_entrypoint.py:55` locks in the new behaviour without noting that it differs from `main`.

Treating `None` as "not passed" is a defensible and arguably better contract; the problem is that it ships undeclared.

**Change Requested:** Keep the `is not None` semantics, and (a) correct the stale claim in the PR description, (b) add the `frame_in_def_cal` / `play_in_cal_block` `None` change to the CHANGELOG entry, and (c) note in the `loads()` docstring that an explicit `None` means "not passed" for every kwarg — the docstring currently gives no way for a reader to predict this.

src/pyqasm/entrypoint.py
```diff
@@ -32,6 +32,37 @@
...
+def _validate_kwargs(kwargs: dict) -> None:
+    """Reject unknown kwarg names and non-positive values at the call site,
+    instead of silently dropping them (issue #356)."""
+    unknown = sorted(set(kwargs) - set(_LOADS_KWARG_ATTRS))
+    if unknown:
+        raise TypeError(f"loads() got unexpected keyword argument(s): {', '.join(unknown)}")
+    for name in _POSITIVE_KWARGS:
+        value = kwargs.get(name)
+        if value is not None and name in kwargs and value <= 0:
+            raise ValueError(f"loads() kwarg '{name}' must be positive, got {value!r}")
```
TheGupta2012: **Type:** Implementation
**Severity:** Medium

**Rationale:** `value <= 0` runs without a type check, which produces two results that work against the PR's stated goal of failing clearly at the call site:

1. Any non-numeric value raises a bare comparison error that names neither the kwarg nor the function:
   ```
   >>> loads(src, device_qubits="5")
   TypeError: '<=' not supported between instances of 'str' and 'int'
   ```
   Same for `{}`, `[]` and `complex(1)`. A caller who reads a device config from JSON and gets `"5"` learns nothing about which argument was wrong. On `main` the value was stored as-is, so this is not a regression — but it is the exact failure mode this validator exists to prevent.

2. `bool` is an `int`, so `True <= 0` is `False` and `loads(src, device_qubits=True)` passes validation and stores `True`. It then behaves as `1` at `visitor.py:502` (`total_qubits > True`). The mirror case, `device_qubits=False`, produces `ValueError: ... must be positive, got False`, which reads as though `False` were a plausible qubit count.

The redundant `name in kwargs` is folded into the suggestion: `kwargs.get(name) is not None` already implies the key is present.

**Change Requested:** Guard the type before comparing, and reject `bool` explicitly.

```suggestion
    for name in _POSITIVE_KWARGS:
        value = kwargs.get(name)
        if value is None:
            continue
        if isinstance(value, bool) or not isinstance(value, (int, float)):
            raise TypeError(f"loads() kwarg '{name}' must be a number, got {type(value).__name__}")
        if value <= 0:
            raise ValueError(f"loads() kwarg '{name}' must be positive, got {value!r}")
```

Worth adding parametrised cases for `"5"`, `[]` and `True` to `tests/test_entrypoint.py` — the current suite covers only `0`, `0.0` and negatives, so none of the above is caught.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Types of changes:
 ### Removed
 
 ### Fixed
+- Fixed `loads()` silently dropping falsy kwarg values (e.g. `extern_functions={}`) by testing truthiness instead of presence. `loads()` now also rejects unknown kwarg names with a `TypeError` and non-positive `device_qubits` / `device_cycle_time` / `compiler_angle_type_size` / `frame_limit_per_port` with a `ValueError`, so mistakes fail at the call site. ([#356](https://github.com/qBraid/pyqasm/issues/356))
```
**Severity:** Medium

**Rationale:** Two separate problems with this entry.

**It does not describe the shipped behaviour.** "Fixed `loads()` silently dropping falsy kwarg values" implies falsy values are now honoured. Measured against `main`, per kwarg:

- `device_qubits=0`, `device_cycle_time=0.0`, `compiler_angle_type_size=0`, `frame_limit_per_port=0` — previously dropped, now **raise `ValueError`**. Not honoured. (Rejecting is the right call: `modules/base.py:628` guards the device-qubit check with `if self._device_qubits:`, so a stored `0` would have been skipped anyway. Issue #356 offers this as an acceptable resolution.)
- `frame_in_def_cal=False`, `play_in_cal_block=False` — already honoured on `main` via the `in kwargs` test. No change.
- `extern_functions={}` / `[]` — the only values genuinely newly stored, and the issue itself calls this case harmless.

So the sole behavioural gain in the "no longer dropped" direction is the one the entry cites parenthetically, while the four cases a reader would most expect now raise. The clause "by testing truthiness instead of presence" also inverts the direction of the fix, and the mechanism is in fact neither: it is `is not None` (`entrypoint.py:140`).

**The breaking part is filed under `Fixed`.** `TypeError` on unknown kwarg names changes `loads(src, devise_qubits=5)` from a silent no-op to a hard failure. Any existing caller passing a stray or misspelled kwarg breaks on upgrade. This repository states it follows Keep a Changelog and SemVer, and `loads()` is the primary public entrypoint, so this belongs where a reader scanning for upgrade risk will find it.

**Change Requested:** Rewrite the entry to state that non-positive values are now *rejected* rather than honoured, name `extern_functions` as the case that is newly stored, drop the inverted "truthiness instead of presence" clause, and add the `frame_in_def_cal=None` / `play_in_cal_block=None` change. Move the unknown-kwarg rejection to `Improved / Modified` with an explicit note that it is a breaking change for callers passing unrecognised kwargs.

src/pyqasm/entrypoint.py
```diff
@@ -74,13 +105,16 @@ def loads(program: openqasm3.ast.Program | str, **kwargs) -> QasmModule:
...
+        TypeError: If the input is not a string or an `openqasm3.ast.Program` instance,
+            or if an unrecognized keyword argument is passed.
+        ValueError: If a numeric keyword argument is zero or negative.
         ValidationError: If the program fails parsing or semantic validation.
 
     Returns:
         QasmModule: An object containing the parsed qasm representation along with
             some useful metadata and methods
     """
+    _validate_kwargs(kwargs)
```
TheGupta2012, line 147: **Type:** Implementation
**Severity:** Low

**Rationale:** `load()` forwards `**kwargs` straight through (`entrypoint.py:82`), so both new exceptions surface from `load()` as well — but its docstring documents neither `**kwargs` nor a `Raises` section, while `loads()` gained both. The messages also hardcode the wrong name: `load("f.qasm", devise_qubits=5)` reports `loads() got unexpected keyword argument(s): devise_qubits`, pointing the caller at a function they did not call.

**Change Requested:** Add a `Raises` section to `load()` covering `TypeError` and `ValueError` and cross-reference `loads()` for the supported `**kwargs`. Optionally pass the caller's name into `_validate_kwargs` (`_validate_kwargs(kwargs, func="load")`) so the message names the entrypoint that was actually invoked.

TheGupta2012 (changes requested):
## Verdict

Request changes — the refactor itself is a clear improvement, but three items should be settled before merge.

Replacing seven hand-written `if` blocks with `_LOADS_KWARG_ATTRS` is the right structural call: the kwarg list, the docstring and the storage loop can no longer drift apart, and rejecting unknown names closes a real silent-typo hole. The concerns are about the edges of the new validator and about the CHANGELOG describing behaviour the code does not have.

## Findings

| # | Type | Severity | Item |
|---|---|---|---|
| 1 | Implementation | Medium | `frame_in_def_cal=None` / `play_in_cal_block=None` silently flip behaviour vs `main` (`entrypoint.py:140`) |
| 2 | Implementation | Medium | `_validate_kwargs` has no type check: bare comparison `TypeError`, and bools pass positivity (`entrypoint.py:63`) |
| 3 | Maintenance | Medium | CHANGELOG entry does not match the shipped behaviour, and the breaking part sits under `Fixed` (`CHANGELOG.md:29`) |
| 4 | Implementation | Low | `load()` inherits the new exceptions but not the docs; both messages name `loads()` (`entrypoint.py:117`) |

**Minor polish, no separate comments needed**

- `entrypoint.py:63` — `value is not None and name in kwargs` is redundant; `kwargs.get(name) is not None` already implies `name in kwargs`. The suggestion on finding 2 drops it.
- `entrypoint.py:136-138` — the comment says "presence tests, not truthiness", but the code tests `is not None`, which is a third thing. Rewording it to name `is not None` would keep the comment true as the file evolves.
- `entrypoint.py:141` — `setattr` with attribute names held as strings means a rename in `modules/base.py` creates a phantom attribute instead of failing. A one-line `hasattr` assert in `_validate_kwargs`, or a module-import-time check over `_LOADS_KWARG_ATTRS.values()`, would catch that.
- Strictness is now uneven across the public API: `QasmModule.unroll(**kwargs)` (`modules/base.py:640`) still forwards unrecognised names to `QasmVisitor` without complaint and still uses walrus truthiness (`external_gates`, `consolidate_qubits`). Harmless today, since both defaults absorb the falsy case, but a caller who learns that `loads()` rejects typos will reasonably assume `unroll()` does too.

## How it was tested

- Behaviour matrix run on this head and on `origin/main` (`7c31308`), for all seven kwargs across omitted / `None` / `0` / `0.0` / `False` / `True` / `{}` / `[]` / normal / `-5` / `"5"` / `complex(1)`. Every claim below comes from that run.
- Full suite in a worktree at `d1a85065`: **729 passed, 2 failed, 4 skipped**. The same 2 `tests/cli/test_cli_commands.py` failures reproduce on `origin/main` (714 passed, same 2 failed) and are an artifact of the long checkout path wrapping the expected filename, not this PR. Net effect of the PR: **+15 passing tests, no regressions**.
- End-to-end check of the `None` flip on a real OpenPulse program (`newframe` inside a `defcal`), unrolled on both trees.
- `black --check` passes on both changed files.
- `gh pr checks 361`: only `Changelog` and `CodeRabbit` ran — **no test workflow executed on this PR**, so the local suite above is the only test evidence.

**Not verified:** `pylint` and `isort` are not present in the available environment, so repo lint beyond `black` was not run.

## Next steps

1. Address findings 1-4.
2. Rebase onto `origin/main`. The branch points at `7e05f4c` and `CHANGELOG.md` conflicts; a rebase restores the `#346` / `#349` / `#365` entries that the current diff appears to delete. No code conflict exists.

src/pyqasm/entrypoint.py
ryanhill1: Agreed on all three counts — keeping `is not None` and declaring it properly, in 81b764d.

(a) The PR description's auto-generated section was flatly wrong and is gone. (b) The changelog entry now spells out that `frame_in_def_cal=None` / `play_in_cal_block=None` mean "not passed", that this differs from `main` where an explicit `None` was stored and read as falsy, and that `False` reaches the old state. (c) The `loads()` docstring now says an explicit `None` means "not passed" for every kwarg and to pass `False` to turn off a boolean.

Your framing is right: nothing is lost in capability, the problem was that it shipped undeclared.

src/pyqasm/entrypoint.py
ryanhill1: Both correct, applied essentially verbatim in 81b764d — including the redundant `name in kwargs`, which is gone.

The `bool` case is the one I would have missed: `device_qubits=True` passing validation and behaving as `1` at `visitor.py:502` is worse than a loud rejection, and `must be positive, got False` reading as though `False` were a plausible qubit count is exactly backwards.

Added the parametrized cases you asked for — `"5"`, `[]`, `complex(1)`, `{}` and both bools.

CHANGELOG.md
ryanhill1: Both points taken — rewritten in 81b764d.

The `Fixed` entry now walks the kwargs individually: `extern_functions={}` / `[]` is the case genuinely newly stored, the four numeric kwargs are **rejected** rather than honoured, and the `frame_in_def_cal` / `play_in_cal_block` `None` change is named. The "truthiness instead of presence" clause is gone; the mechanism is described as `is not None`, which is what the code does.

The unknown-kwarg `TypeError` moved to `Improved / Modified` prefixed **Breaking**, spelling out that a caller passing a stray or misspelled kwarg breaks on upgrade. You are right that it does not belong under `Fixed` in a repo that claims Keep a Changelog and SemVer for its primary entrypoint.

Also rebased onto `origin/main` — the `#346` / `#349` / `#365` entries are back, and it was a changelog-only conflict as you predicted.

src/pyqasm/entrypoint.py
ryanhill1: Fixed in 81b764d. `_validate_kwargs` takes a `func` name, and `load()` calls it directly before `process_include_statements` so the message names the entrypoint the caller invoked. `load()` also gains the `Raises` section and a `**kwargs` pointer to `loads()`.

Pinned with `test_load_errors_name_load_not_loads`, covering both the `TypeError` and the `ValueError`.

TheGupta2012 (approved):

PR #362, opened 2026-08-07
fix: reject mixed declared/physical programs under consolidate_qubits

### LINKED ISSUE #353: consolidate_qubits=True emits two address spaces for mixed declared/physical programs
## Description

For a program that mixes declared registers with physical qubits, `unroll(consolidate_qubits=True)` emits OpenQASM describing **two address spaces**, while `num_qubits` collapses them into one. Nothing in the output states which is intended.

```
qubit[2] q;                      qubit[2] __PYQASM_QUBITS__;
h q[0];               ------>    h __PYQASM_QUBITS__[0];
cx q[0], q[1];                   cx __PYQASM_QUBITS__[0], __PYQASM_QUBITS__[1];
h $1;                            h $1;
```

`module.num_qubits == 2`. Internally `$1` and `__PYQASM_QUBITS__[1]` are counted as the same qubit — `_register_physical_qubit` does `num_qubits = max(num_qubits, phys_idx + 1)` — but the emitted program never says so. A consumer sees a virtual register the hardware may map anywhere, plus an absolute reference to physical qubit 1. If they are the same qubit, the output should say so; if they are not, `num_qubits` under-counts.

A second shape makes it starker:

```
qubit[2] q;     ------>    qubit[6] __PYQASM_QUBITS__;
h q[0];                    h __PYQASM_QUBITS__[0];
h $5;                      h $5;
```

`num_qubits == 6` — a six-slot register of which one slot is used, sized by a physical index that register does not contain.

## Related: unreferenced consolidated declaration

For a program with no declared registers at all, the consolidated declaration is still emitted and nothing references it:

```
h $1;    ------>    qubit[2] __PYQASM_QUBITS__;
                    h $1;
```

It parses, but a device backend may allocate against a register the program never uses. Suppressing it when nothing was consolidated — `sum(global_qreg_size_map.values()) == 0` in `_add_pyqasm_qubit_register` — would fall out of whichever option below is chosen.

Note that with `device_qubits` set, the declaration is sized from that instead, giving a third number: `loads(src, device_qubits=5)` on `h $1; cz $2, $1;` emits `qubit[5] __PYQASM_QUBITS__;` while `num_qubits == 3`.

## Provenance

Pre-existing #325 semantics, not introduced by #344. Running both shapes on `main` through the `measure`/`reset` paths (`qubit[2] q; h q[0]; c = measure $1;` and `qubit[2] q; h q[0]; reset $5;`) gives identical output and identical counts. #344 only stops the `AttributeError` that previously masked the question for gate operands.

## Options

Author's call — this issue exists to settle it rather than to prescribe:

1. **Map physical operands into the consolidated register** using the same offset scheme, so the whole program speaks one address space — arguably what "consolidate" implies.
2. **Raise a `ValidationError`** when a program mixes declared registers with physical qubits under `consolidate_qubits=True`, rather than emitting something ambiguous.
3. **Keep current behaviour**, document the semantics explicitly, and fix `num_qubits` so it does not conflate the two address spaces.

`tests/qasm3/test_device_qubits.py::test_physical_qubits_only` and `::test_physical_qubits_are_not_consolidated` assert `num_qubits` explicitly, so whichever option is taken shows up as a deliberate test edit.

Found in review of #344 (M1, L1).

ryanhill1:
Fixes #353

Settles the semantics question from the issue as **option 2**: a program mixing declared registers with physical qubits under `unroll(consolidate_qubits=True)` raises a `ValidationError` naming the physical qubits, rather than emitting a consolidated register plus as-written `$n` references — two address spaces the output never relates. This preserves #344's "physical qubits are left as written" rule for pure-physical programs while refusing to emit an ambiguous program.

Falling out of the same check: a program using only physical qubits no longer receives a `__PYQASM_QUBITS__` declaration nothing references (previously sized by `device_qubits` or the highest physical index).

The two #344 tests that pinned the old counts (`test_physical_qubits_are_not_consolidated`, `test_physical_qubits_only`) are deliberately edited, as the issue anticipated.

**Also in this PR:**
- Updates the changelog to document the consolidated-qubit validation behavior.

<details>
<summary>Consolidated Qubit Validation Flow</summary>

```mermaid
sequenceDiagram
  participant Test as tests/qasm3/test_device_qubits.py
  participant Visitor as src/pyqasm/visitor.py ⚠️
  Test->>Visitor: unroll(consolidate_qubits=True)
  Visitor->>Visitor: inspect declared and physical qubits
  Visitor-->>Test: raise ValidationError or emit unrolled program
```
</details>

<sub>Auto-enriched by [Argus](https://argus.reviews)</sub>


ryanhill1:
@argus-eye review

ryanhill1:
@argus-eye review

src/pyqasm/visitor.py
```diff
@@ -496,15 +496,33 @@ def _qubit_register_consolidation(
...
+            # presence, not capacity: a zero-sized declared register still declares
+            # a second address space
+            if self._global_qreg_size_map:
+                raise_qasm3_error(
+                    "Cannot consolidate qubit registers: the program mixes declared "
+                    f"registers with physical qubits ({', '.join(physical_qubits)})",
+                )
+            # only physical qubits: nothing to consolidate, so do not declare an
+            # internal register nothing would reference
+            return unrolled_stmts
```
TheGupta2012, line 551: **Type:** Implementation
**Severity:** Medium

**Rationale:** This early return skips the `INTERNAL_QUBIT_REGISTER` reserved-name loop directly below, and the check is reachable with that name already declared. Verified against `origin/main`:

| program | `origin/main` | this PR |
|---|---|---|
| `int __PYQASM_QUBITS__ = 3; h $1;` | `ValidationError: Variable '__PYQASM_QUBITS__' is already defined` | succeeds silently |
| `qubit[2] __PYQASM_QUBITS__; h $1;` | same reserved-name error | `...mixes declared registers with physical qubits ($1)` |

Neither outcome corrupts the output — no internal register is emitted on this path, so nothing actually collides. The concern is the contract: the docstring edited a few lines above still promises a raise "if the reserved register `'__PYQASM_QUBITS__'` is already declared", and on this path it no longer does. The second row is also a diagnostic downgrade: the user's real problem is the reserved name, but they are told about physical qubits instead.

**Change Requested:** Move the `global_scope` reserved-name loop (lines 526-532) above the new physical-qubit block, so it runs before either exit. One move fixes both rows and keeps the docstring honest. If the intent is instead that the guard should not apply when no register is emitted, narrow the docstring to say so.

src/pyqasm/visitor.py
```diff
@@ -496,15 +496,33 @@ def _qubit_register_consolidation(
...
             raise_qasm3_error(
                 # pylint: disable-next=line-too-long
                 f"Total qubits '({total_qubits})' exceed device qubits '({self._module._device_qubits})'.",
             )
 
+        # physical qubits are kept as written, so consolidating around them would emit
+        # two address spaces the output cannot relate (issue #353)
+        physical_qubits = sorted(
+            name for name, _ in self._module._qubit_depths if name.startswith("$")
+        )
```
TheGupta2012: **Type:** Implementation
**Severity:** Low

**Rationale:** `sorted()` on the raw `$n` strings sorts lexicographically. A program using `$2` and `$10` produces `...physical qubits ($10, $2)`, verified at this commit. On a device with more than ten qubits the list reads as unordered, which makes it harder to scan for the offending references.

**Change Requested:** Sort by the numeric index.

```suggestion
        physical_qubits = sorted(
            (name for name, _ in self._module._qubit_depths if name.startswith("$")),
            key=lambda name: int(name[1:]),
        )
```

The parentheses around the generator are required once `key=` is added. Every name reaching this point has already passed the `reg_name[1:].isdigit()` guard in `_get_op_bits`, so `int()` is safe.

src/pyqasm/visitor.py
```diff
@@ -496,15 +496,33 @@ def _qubit_register_consolidation(
...
+        if physical_qubits:
+            # presence, not capacity: a zero-sized declared register still declares
+            # a second address space
+            if self._global_qreg_size_map:
+                raise_qasm3_error(
+                    "Cannot consolidate qubit registers: the program mixes declared "
+                    f"registers with physical qubits ({', '.join(physical_qubits)})",
```
TheGupta2012: **Type:** Maintenance
**Severity:** Low

**Rationale:** This is a new hard failure on a public flag: programs that unrolled successfully before now raise. The message names the physical qubits but does not tell the user how to proceed, and `raise_qasm3_error` is called without `error_node`/`span` here — no statement node is available at `finalize` time — so there is no line number either. The message text is the entire diagnostic.

**Change Requested:** Name the two ways out.

```suggestion
                    "Cannot consolidate qubit registers: the program mixes declared "
                    f"registers with physical qubits ({', '.join(physical_qubits)}). "
                    "Unroll without 'consolidate_qubits=True', or rewrite the physical "
                    "qubits as operands of a declared register.",
```

TheGupta2012 (commented):
## Verdict: approve with minor comments

Settling #353 as option 2 is a defensible call, and the implementation is placed correctly: the check sits in `_qubit_register_consolidation` after the `device_qubits` capacity check, so the existing "total qubits exceed device qubits" error still wins when both apply. The scan over `_qubit_depths` turns out to be complete — every path that can put a physical qubit into a program registers it. No blocking issues. Four comments below: one Medium about a validation that the early return skips (and which the docstring in this PR still promises), three Low.

### Findings

- [ ] **Medium** — the `return unrolled_stmts` on the pure-physical path skips the reserved-name check, contradicting the docstring edited in this same PR. `int __PYQASM_QUBITS__ = 3; h $1;` raises `Variable '__PYQASM_QUBITS__' is already defined` on `origin/main` and succeeds silently here. Hoisting the loop above the new block also restores the precise message for the qreg shape.
- [ ] **Low** — `sorted()` on `$n` strings orders `$10` before `$2` in the error message.
- [ ] **Low** — the error names the offending qubits but does not say what to do next, and no source span is available at `finalize` time, so the message text is all the user gets.
- [ ] **Low** — `test_mixed_declared_and_physical_still_unrolls_without_consolidation` asserts only `num_qubits`; the deleted test pinned the emitted QASM, so nothing now pins #344's "left as written" rule for a mixed program.

### How it was tested

Behaviour was compared between a worktree at `c18b624` and one at `origin/main` (`7c31308`), running every claim rather than reading for it.

**Coverage of the physical-qubit scan (complete).** A physical qubit reaching the program only through `measure`, `reset`, `barrier`, an `if` body, a `box` body, or a custom gate call site all land in `_qubit_depths` — `_get_op_bits` registers `$n` for every operand shape. All six now raise; all six emitted the two-address-space output on base.

**No false positive on the pure-physical path.** `_global_qreg_size_map` holds only user-declared registers at that point, so `h $1; cz $2, $1;` with `consolidate_qubits=True` returns cleanly, `num_qubits == 3`, no `__PYQASM_QUBITS__` — with and without `device_qubits`.

**`device_qubits` ordering is right.** `h $9;` under `device_qubits=3` still raises `Total qubits '(10)' exceed device qubits '(3)'.` — the capacity check fires before the new block, for physical-only and mixed programs alike.

**"Presence, not capacity" holds.** `qubit[0] q;` does create a `_global_qreg_size_map` entry, so the truthiness test is correct. (`qreg q[0];` is unreachable — the QASM 2 parser rejects a zero-size `qreg` outright, so the qasm3 spelling is the only one that can exercise this.)

**No regression with the flag off.** The mixed program unrolls unchanged under plain `unroll()`. A corpus of 333 declared-registers-only programs harvested from `tests/` dumps byte-identically on both worktrees under `consolidate_qubits=True`; the only two diffs are set-iteration order inside an unrelated "Unsupported OpenQASM version" message.

**OpenPulse does not trip the new check.** Worth recording, because `defcalgrammar "openpulse"` forces `_consolidate_qubits = True` at `visitor.py:3328` regardless of what the caller passed. Under OpenPulse, gate, measure and reset operands are renamed to `__PYQASM_QUBITS__[n]` before registration, so no `$n` key ever reaches `_qubit_depths` and plain `unroll()` on an OpenPulse program behaves identically on both worktrees.

**Suite and CI.** 720 passed / 714 on base (the 6 new tests), same 2 pre-existing `tests/cli` failures on both worktrees, so nothing is attributable to this change. `black --check` clean on both touched files; all GitHub checks green. `isort` and `pylint` are absent from the local venv and were not run — CI covers them.

**Blast radius.** An org-wide code search finds no consumer of `consolidate_qubits` outside pyqasm itself and the docs, so the breaking change lands narrowly. That search does not see private repositories.

### Next steps

1. Rebase onto `origin/main` and resolve the `src/pyqasm/visitor.py` conflict against #346's typing and docstring changes. #357 and #359 touch other regions of the same file.
2. `qBraid/docs` `v2/pyqasm/user-guide/advanced-features.mdx` documents `consolidate_qubits` without the new rejection. Its examples use declared registers only, so they still hold, but the page should gain a line about mixed programs.
3. Two pre-existing behaviours surfaced while testing, identical on `origin/main` and therefore not findings against this PR, but each worth its own issue:
   - Under OpenPulse, `barrier $0` is the one physical reference left unrenamed while every other one becomes `__PYQASM_QUBITS__[n]`, so the output carries both spellings. Index-identity makes it unambiguous in practice, only inconsistent.
   - Unrolling a module twice resets `num_qubits` from 3 to 2 for a mixed program, dropping the physical qubit from the count. This PR makes the path easy to reach: catching the new error and re-running `unroll()` without the flag is the natural remedy, and it lands on the stale count.

tests/qasm3/test_device_qubits.py
```diff
@@ -331,48 +331,84 @@ def test_incorrect_qubit_reg(qasm_code, error_message, error_span, caplog):
...
+    h q[0];
+    cz $2, q[1];
     """
     result = loads(qasm, device_qubits=5)
-    result.unroll(consolidate_qubits=True)
-    check_unrolled_qasm(dumps(result), expected_qasm)
-    # two consolidated slots plus physical $2, which sizes the count to its own index + 1.
-    # neither number is the declared qubit[5], which comes from device_qubits (see #353)
+    result.unroll()
     assert result.num_qubits == 3
```
TheGupta2012, line 436: **Type:** Implementation
**Severity:** Low

**Rationale:** The `test_physical_qubits_are_not_consolidated` this replaces asserted the full unrolled text, which is what pinned #344's rule that a physical qubit survives unrolling as written. This test asserts only the count, and `num_qubits == 3` would still pass if `cz $2, q[1]` came out rewritten or dropped. `test_physical_qubits_only` still pins the text for the pure-physical case, so the gap is specifically the mixed shape.

**Change Requested:** Add the output assertion. Verified against this commit — `check_unrolled_qasm` and `dumps` are already imported in this file.

```suggestion
    result.unroll()
    expected_qasm = """OPENQASM 3.0;
    include "stdgates.inc";
    qubit[2] q;
    h q[0];
    cz $2, q[1];
    """
    check_unrolled_qasm(dumps(result), expected_qasm)
    assert result.num_qubits == 3
```

src/pyqasm/visitor.py
ryanhill1: Reproduced both rows exactly, and took your first option in f4fcfc6 — the reserved-name loop now runs before either physical-qubit exit. Keeping the guard is the right reading: the docstring promise stays true, and a user who declared `__PYQASM_QUBITS__` is told about that rather than about physical qubits.

Pinned with a parametrized test over both the `int` and `qubit[2]` declarations.

src/pyqasm/visitor.py
ryanhill1: Confirmed — `($10, $2)` on a program using `$2` and `$10`. Applied your suggestion verbatim in f4fcfc6, and added a test that would catch a regression to lexicographic order.

src/pyqasm/visitor.py
ryanhill1: Applied in f4fcfc6. Your point about the message being the entire diagnostic is the one that convinced me — no statement node exists at `finalize` time, so there is no span to fall back on, and telling someone their program is rejected without telling them what to do about it is a poor trade for a new hard failure on a public flag. Added a test asserting the way-out text is present.

tests/qasm3/test_device_qubits.py
ryanhill1: Fair — the count assertion alone would not have caught `cz $2, q[1]` being rewritten or dropped, which is the whole #344 rule the old test existed to pin. Added the output assertion in f4fcfc6.

TheGupta2012 (approved):

Issue #363, opened 2026-08-07
review and fix deprecated typing aliases


micpap25:
Codebase uses deprecated typing imports.

Ex. 1: `ast.py` L14: `from typing import List, Optional, Union, Tuple`
`List` and `Tuple` could be replaced with built-ins, and arguably `Union` could be replaced with `|` and `Optional` could be reveiwed.

Ex. 2: `base.py` L25: `from typing import Iterator, Optional, Sequence, TypeVar`
`Iterator` and `Sequence` should be imported from `collections.abc`.

PR #365, opened 2026-08-08
Update pulse/visitor.py

### LINKED ISSUE #364: update docstrings and type hinting in pulse/visitor.py
Follow-up to #346
As discussed, sweep `pulse/visitor.py` to standardize the docstring in relation to the updated `visitor.py`.

micpap25:
Fixes #364 
As a follow-up to #347 and #363, apply similar changes to `pulse.visitor.py`.

### DIFF
--- a/src/pyqasm/pulse/visitor.py
+++ b/src/pyqasm/pulse/visitor.py
@@ -104,7 +104,13 @@ class OpenPulseVisitor:
 
         Args:
             statement (Any): The statement that is calling the function.
-            frame (str): The frame to get.
+            frame (str): The name of the frame to get.
+
+        Returns:
+            Frame: The frame that was gotten.
+
+        Raises:
+            ValidationError: If there is no frame with the provided name.
         """
         frame_obj = self._openpulse_scope_manager.get_from_visible_scope(frame)
         if not isinstance(frame_obj, Frame) or frame_obj is None:
@@ -134,6 +140,9 @@ class OpenPulseVisitor:
             statement (Any): The statement that is calling the function.
             identifier (Identifier | str): The identifier to get the value of,
                 or a string with its name.
+
+        Returns:
+            Any: The value of the identifier.
         """
         if isinstance(identifier, str):
             identifier = qasm3_ast.Identifier(name=identifier)
@@ -338,7 +347,7 @@ class OpenPulseVisitor:
 
         waveform_validators[wf_func_name]()
 
-    def _handle_scope_sync(self, statement: Any, return_value: Optional[Any]) -> None:
+    def _handle_scope_sync(self, statement: Any, return_value: Any = None) -> None:
         """Handle synchronization between QASM and OpenPulse scopes.
 
         Args:
@@ -372,7 +381,7 @@ class OpenPulseVisitor:
             statement (QuantumBarrier): The barrier statement to visit.
 
         Returns:
-            list[QuantumBarrier]: The list containing a single multi-qubit barrier statement.
+            list[QuantumBarrier]: The list containing the original barrier statement.
         """
         if barrier.qubits:
             for qubit in barrier.qubits:
@@ -664,7 +673,7 @@ class OpenPulseVisitor:
 
     def _visit_function_call(  # pylint: disable=too-many-branches, too-many-statements
         self, statement: qasm3_ast.FunctionCall
-    ) -> tuple[Any, list[qasm3_ast.FunctionCall]]:
+    ) -> tuple[Any, list[qasm3_ast.Statement | qasm3_ast.FunctionCall]]:
         """Visit a function call element.
 
         Args:
--- a/src/pyqasm/pulse/visitor.py
+++ b/src/pyqasm/pulse/visitor.py
@@ -103,8 +103,8 @@ class OpenPulseVisitor:
         """Get the frame from the global scope.
 
         Args:
-            frame (str): The frame to get.
             statement (Any): The statement that is calling the function.
+            frame (str): The frame to get.
         """
         frame_obj = self._openpulse_scope_manager.get_from_visible_scope(frame)
         if not isinstance(frame_obj, Frame) or frame_obj is None:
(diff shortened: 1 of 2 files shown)


src/pyqasm/pulse/visitor.py
```diff
@@ -653,14 +664,15 @@ def _visit_classical_declaration(  # pylint: disable=too-many-branches, too-many
 
     def _visit_function_call(  # pylint: disable=too-many-branches, too-many-statements
         self, statement: qasm3_ast.FunctionCall
-    ) -> tuple[Any, list[qasm3_ast.Statement | qasm3_ast.FunctionCall]]:
+    ) -> tuple[Any, list[qasm3_ast.FunctionCall]]:
```
ryanhill1: **D1 — this narrowing is incorrect, and mypy is structurally unable to catch it.**

The function ends with:

```python
statement = qasm3_ast.ExpressionStatement(expression=statement)  # type: ignore
return _return_value, [statement]
```

So the returned list holds an `ExpressionStatement` (a `Statement`, not a `FunctionCall`). The old annotation `tuple[Any, list[Statement | FunctionCall]]` was right. `mypy src` passes only because the `# type: ignore` on the assignment makes mypy believe `statement` is still a `FunctionCall` afterwards — the one situation where a green mypy run doesn't validate the annotation.

Your own new docstring says `tuple[Any, list[Statement | FunctionCall]]`, contradicting the signature two lines above it. Revert the signature; keep the docstring.

src/pyqasm/pulse/visitor.py
```diff
@@ -361,10 +369,10 @@ def _visit_barrier(self, barrier: qasm3_ast.QuantumBarrier) -> list[qasm3_ast.Qu
         """Visit a barrier statement element.
 
         Args:
-            statement (qasm3_ast.QuantumBarrier): The barrier statement to visit.
+            statement (QuantumBarrier): The barrier statement to visit.
 
         Returns:
```
ryanhill1, line 383: **R1** — wording borrowed from `QasmVisitor._visit_barrier`, which consolidates into a single multi-qubit barrier. This one validates and returns the original barrier unchanged (which may be single-qubit). "The list containing the original barrier statement." is accurate.

src/pyqasm/pulse/visitor.py
```diff
@@ -329,13 +338,12 @@ def _check_waveform_functions(
 
         waveform_validators[wf_func_name]()
 
-    def _handle_scope_sync(self, statement: Any, return_value: Any = None) -> None:
+    def _handle_scope_sync(self, statement: Any, return_value: Optional[Any]) -> None:
```
ryanhill1: **N1** — this drops the `= None` default: an API change in a docs sweep. Both current call sites (L658, L661) pass two arguments, so nothing breaks today — but nothing was gained either, and `Optional[Any]` is redundant (`Any` already admits `None`; mypy treats them identically). Either restore `return_value: Any = None` or keep this — just flagging that it's behavioural surface, not docs.

src/pyqasm/pulse/visitor.py
```diff
@@ -103,8 +103,8 @@ def _get_frame(self, statement: Any, frame: str) -> Frame:
         """Get the frame from the global scope.
 
         Args:
-            frame (str): The frame to get.
             statement (Any): The statement that is calling the function.
```
ryanhill1, line 106: **N2** — you're editing this docstring and the signature returns `Frame`, but there's still no `Returns:` section. Since #364 is a sweep, worth adding: `Frame: The frame object from the visible scope.`

src/pyqasm/pulse/visitor.py
```diff
@@ -131,8 +131,9 @@ def _get_identifier(self, statement: Any, identifier: qasm3_ast.Identifier | str
         """Get the value of an identifier.
 
         Args:
-            identifier (qasm3_ast.Identifier): The identifier to get the value of.
             statement (Any): The statement that is calling the function.
```
ryanhill1, line 140: **N2 (cont.)** — same here: signature returns `Any` (the identifier's variable object), no `Returns:` section. These two are the only remaining doc/signature gaps in the file after this PR — adding both makes the #364 sweep genuinely complete.

ryanhill1 (commented):
Thanks for closing the loop on #364 so quickly — checked out and verified at `700678b`. **Toolchain is fully green on this head: 696 passed, 3 skipped; black, isort, mypy clean; pylint 10.00/10.** The Args reorderings, the `_check_identifier` docstring, and the `Returns:` fixes for `_visit_barrier`/`_visit_classical_*` all check out against the actual return statements.

One real defect and three small items, labelled inline:

| | |
|---|---|
| **D1** | `_visit_function_call` return annotation narrowed *incorrectly* — mypy can't see it because of a pre-existing `# type: ignore` |
| **R1** | `_visit_barrier` Returns wording borrowed from the wrong class |
| **N1** | `_handle_scope_sync` lost its default value — benign today, but an API change in a docs sweep |
| **N2** | two methods this PR edits still have no `Returns:` section (`_get_frame`, `_get_identifier`) |

D1 is the only one I'd hold the merge for — the annotation is now wrong in a way the type checker is structurally unable to catch.

src/pyqasm/pulse/visitor.py
micpap25: My bad, mistakenly believed my change was preferred by Python standards but it is not.

ryanhill1 (commented):
Re-checked at `2c7f0ea`. **All five items fixed** — resolved the threads.

- **D1** — signature back to `tuple[Any, list[Statement | FunctionCall]]`, matching both the code and the docstring.
- **R1, N1, N2** — all exactly as suggested; the bonus `Raises:` on `_get_frame` is a nice touch.

Toolchain on this head: **696 passed, 3 skipped; black, isort, mypy clean; pylint 10.00/10.** My doc-vs-signature sweep over `pulse/visitor.py` now reports **zero** gaps — #364's sweep is genuinely complete. Good to go from my side.

ryanhill1 (approved):

PR #371, opened 2026-08-16
Fix while_loop hint, update pragma hint in visitor.py

micpap25:
## Summary of changes

Fixes a mistake introduced by #346, which claims that `visitor:_visit_while_loop` is affected by `self._check_only`.
Additionally, updates the docstring of `visitor:_visit_pragma` to match the rest of the file.

### DIFF
--- a/src/pyqasm/visitor.py
+++ b/src/pyqasm/visitor.py
@@ -2650,8 +2650,7 @@ class QasmVisitor:
             statement (WhileLoop): The while-loop AST node.
 
         Returns:
-            list[Statement]: The list of unrolled statements from the while-loop, or an
-                empty list if self._check_only is True.
+            list[Statement]: The list of unrolled statements from the while-loop.
 
         Raises:
             ValidationError: If loop condition is non-classical or dynamic.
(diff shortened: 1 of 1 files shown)


src/pyqasm/visitor.py
```diff
@@ -3047,9 +3046,10 @@ def _visit_pragma(self, statement: qasm3_ast.Pragma) -> list[qasm3_ast.Pragma]:
         still copied to the output, where the consumer applies it to whatever comes next.
 
         Args:
-            statement (qasm3_ast.Pragma): The Pragma node to visit.
+            statement (Pragma): The Pragma node to visit.
```
TheGupta2012, line 3049: This should be left as is, [see the other methods](https://github.com/qBraid/pyqasm/pull/371/changes#diff-802ca57af29fc92ca8eb2f536fb5217c61c874c4c35bd39b7fd82512ee0e9a84R3031)

TheGupta2012 (commented):
Added one comment, but should be okay to merge

src/pyqasm/visitor.py
micpap25: I don't agree, see for example `_visit_include` specifically for use of `Pragma`; to my understanding for this file we are dropping the `qasm3_ast` in docstrings.

ryanhill1: Yeah I agree with @micpap25, the `qasm3_ast` prefix does not belong in the doc-string, because in the sphinx build it will be rendered as-is without the alias substitution. It should either be just `Pragma`, or it should be the full import path.

ryanhill1:
I am moving to approve, @TheGupta2012 could you take a look at my comment and reconsider

ryanhill1 (approved):

PR #374, opened 2026-08-18
docs: reset the changelog after the v1.1.0 release

TheGupta2012:
## Summary

v1.1.0 shipped on 17 August 2026 and carried every entry that was sitting under `## Unreleased`. Compare the [v1.1.0 release notes](https://github.com/qBraid/pyqasm/releases/tag/v1.1.0) against `CHANGELOG.md` on `main`: the two lists are identical. The `Unreleased` section therefore describes released work.

This PR resets it.

## What changed

- Emptied every `Unreleased` subsection (`Added`, `Improved / Modified`, `Fixed`, `Dependencies`). The 23 removed lines are exactly the entries reproduced in the v1.1.0 release notes.
- Added `v1.1.0` to the top of **Past Release Notes**, matching how v1.0.4 and earlier are archived.

Net: `CHANGELOG.md`, +1 / -23. No other file is touched.

## Why now

The v1.0.3 and v1.0.4 releases each left their entries in `Unreleased`, which is what #372 had to clean up after the fact. Resetting at release time keeps that from recurring.


TheGupta2012:
Merging as just docs are updated

Issue #369, opened 2026-08-15
Indirect gate recursion (a -> b -> a) raises RecursionError instead of ValidationError


ryanhill1:
### Environment

- **PyQASM version**: 1.0.4 (`main` @ 5dcae69)
- **Python version**: 3.11
- **Operating system**: macOS

### What happened?

pyqasm detects *direct* gate self-recursion and raises a clean `ValidationError`, but an *indirect* cycle between two gate definitions is not detected and instead exhausts the Python stack.

```python
import pyqasm

# direct — handled correctly
pyqasm.loads('''OPENQASM 2.0;
include "qelib1.inc";
gate a(t) q { a(t) q; }
qreg q[1];
a(0.5) q[0];
''').validate()
# ValidationError: Recursive definitions not allowed for gate 'a'

# indirect — stack blowout
pyqasm.loads('''OPENQASM 2.0;
include "qelib1.inc";
gate a(t) q { b(t) q; }
gate b(t) q { a(t) q; }
qreg q[1];
a(0.5) q[0];
''').validate()
# RecursionError: maximum recursion depth exceeded while calling a Python object
```

For contrast, a forward reference to a gate that is never defined is reported correctly:

```
ValidationError: Unsupported / undeclared QASM operation: b
```

So the three cases behave as: direct cycle → good error; undefined gate → good error; indirect cycle → `RecursionError`.

### Cause

The guard at `visitor.py:1271` compares a single level:

```python
if isinstance(gate_op, qasm3_ast.QuantumGate) and gate_op.name.name == gate_name:
    raise_qasm3_error(
        f"Recursive definitions not allowed for gate '{gate_name}'",
        ...
```

It only catches a body that references its own name. A cycle of length ≥ 2 passes the check and recurses through gate expansion until the interpreter's stack limit is reached.

### Expected behaviour

A `ValidationError` identifying the cycle, in the same shape as the existing direct-recursion message — ideally naming the path, e.g. `Recursive definitions not allowed for gate 'a' (a -> b -> a)`. Tracking the expansion chain and checking membership rather than comparing one name would cover both cases with the same code.

### Why this came up

[...]

PR #375, opened 2026-08-18
fix: nested external gate depth, and gate definition cycles of any length

### LINKED ISSUE #367: Nested external custom gate counts the depth of the decomposition it skips
### Description

A nested external custom gate counts the depth of the decomposition it skipped. This is the same symptom [#352](https://github.com/qBraid/pyqasm/issues/352) fixed for external *basic* gates, in the one shape that fix did not reach.

Raised by @TheGupta2012 while reviewing #359, where the changelog wording implied the custom-gate path was already correct in general. It is correct for a single level of nesting, which is why the existing tests pass.

### Reproduction

```python
from pyqasm import loads

qasm = """OPENQASM 3.0;
include "stdgates.inc";
gate inner a, b { crz(0.5) a, b; }
gate outer a, b { inner a, b; h a; }
qubit[2] q;
outer q[0], q[1];
"""

module = loads(qasm)
module.unroll(external_gates=["outer"])
print(module.depth())  # 13, but exactly one statement is emitted
```

Measured on `7c31308`:

| `external_gates` | statements emitted | `depth()` | expected |
|---|---|---|---|
| `["outer"]` | 1 | **13** | 1 |
| `["inner", "outer"]` | 1 | **2** | 1 |
| `["g"]` (single level, for contrast) | 1 | 1 | 1 |

13 is the fully-decomposed depth — the `crz` decomposition plus the `h`. So an external gate that emits a single statement reports the cost of the body it never emitted.

### Cause

`_visit_custom_gate_operation` assigns `_recording_ext_gate_depth` unconditionally at [`visitor.py:1309`](https://github.com/qBraid/pyqasm/blob/main/src/pyqasm/visitor.py#L1309) and clears it unconditionally at [`visitor.py:1341-1342`](https://github.com/qBraid/pyqasm/blob/main/src/pyqasm/visitor.py#L1341), with no save-restore:

```python
self._recording_ext_gate_depth = self._in_verbatim_box or gate_name in self._external_gates
...
if self._recording_ext_gate_depth:
    self._recording_ext_gate_depth = False
```

With `external_gates=["outer"]`, visiting `outer` sets the flag to `True`. Descending into `inner` overwrites it with `False`, because `inner` is not itself external — so the `crz` decomposition inside records its depth, and when `outer` finishes the flag is already `False`, so `outer` never records its own. The inner gate clobbers the outer gate's state in both directions.

### Suggested fix

Two changes, both in `_visit_custom_gate_operation`. #359 already applies the save-restore half of this pattern in `_visit_external_gate_operation`.

1. Save the flag and restore it on exit, rather than clearing it to `False`.
2. Stay suppressed inside an enclosing external gate — `prev or is_external` rather than a bare assignment — and record only from the outermost external gate.

```python
prev_recording = self._recording_ext_gate_depth
is_external = self._in_verbatim_box or gate_name in self._external_gates
self._recording_ext_gate_depth = prev_recording or is_external
...
self._recording_ext_gate_depth = prev_recording
if is_external and not prev_recording:
    ...
```

I prototyped this locally: all three rows above report `1`, and the full suite passes (713 passed, 3 skipped) with no expectation changes. Note that the save-restore alone is not sufficient — it fixes `["inner", "outer"]` but leaves `["outer"]` at 13, because the nested gate still re-enables recording for the body its parent is skipping.

### Test coverage

`tests/qasm3/test_depth.py` covers single-level external custom gates only. Worth adding the nested shape alongside `test_external_basic_gate_counts_own_depth`, parametrized over `["outer"]` and `["inner", "outer"]`, since the two failed differently.

TheGupta2012:
Fixes #367 and #369.

> **Scope note:** #376 (issue #369, cyclic gate definitions) was merged into this branch, so this PR now carries both fixes plus a small follow-up that inlines #376's single-caller error helper. Both touch `_visit_custom_gate_operation`, which is why they travel together.

## The bug

An external custom gate whose body calls **another custom gate** reported the depth of the decomposition it never emitted. This is the shape the #352 fix did not reach — one level of nesting was already correct, which is why the existing tests passed.

```python
qasm = '''OPENQASM 3.0;
include "stdgates.inc";
gate inner a, b { crz(0.5) a, b; }
gate outer a, b { inner a, b; h a; }
qubit[2] q;
outer q[0], q[1];
'''
module = loads(qasm)
module.unroll(external_gates=["outer"])
module.depth()   # 13, from one emitted statement
```

| `external_gates` | statements | `depth()` before | after |
|---|---|---|---|
| `["outer"]` | 1 | **13** | 1 |
| `["inner", "outer"]` | 1 | **2** | 1 |
| single level, for contrast | 1 | 1 | 1 |
| `None` (nothing external) | 13 | 13 | 13 |

## Cause

`_visit_custom_gate_operation` assigned `_recording_ext_gate_depth` unconditionally and cleared it to `False` on exit, with no save-restore. The inner gate clobbered the outer gate's state in **both** directions:

- Descending into `inner` overwrote the flag with `False`, so the `crz` decomposition inside the body `outer` was skipping recorded its depth → 13.
- When `inner` returned it left the flag `False`, so `outer` never recorded its own single depth → the `["inner", "outer"]` row reported 2 instead of 1.

That is why save-restore alone is not sufficient: it fixes `["inner", "outer"]` and leaves `["outer"]` at 13.

## The fix

Two changes in `_visit_custom_gate_operation`:

1. Save the flag and restore it in a `finally` block, matching the pattern #359 introduced in `_visit_external_gate_operation`.
2. Compute `prev_recording or is_external` rather than assigning, so an inner gate stays suppressed inside an enclosing external gate, and record the depth under `is_external and not prev_recording` — from the outermost external gate only.

## Verification

- Full suite: **789 passed, 4 skipped**. No existing expectation changed.
- **Mutation-tested.** With the source fix reverted, both new parametrizations fail, and they fail *differently* — `["outer"]` at 13, `["inner", "outer"]` at 2 — which is the point: they pin two distinct halves of the bug.

## Tests added

In `tests/qasm3/test_depth.py`, beside `test_external_basic_gate_counts_own_depth`:

- `test_nested_external_custom_gate_counts_own_depth`, parametrized over `["outer"]` and `["inner", "outer"]`, asserting one emitted statement and `depth() == 1`.
- `test_nested_custom_gate_depth_unchanged_without_external_gates` — the same program still decomposes to 13 when nothing is external, so the flag handling cannot silently suppress ordinary counting.
- `test_inner_external_gate_records_depth_inside_plain_custom_gate` — `external_gates=["inner"]` still records the inner gate's own depth when the enclosing gate is not external (2, not 1).

The last two pass on `main`; they are regression guards, not bug pins.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recursive custom-gate validation to detect indirect cycles of any length and report the complete cycle path.
  - Corrected nested external-gate depth tracking, ensuring operation counts reflect the appropriate outermost and inner gates.
  - Preserved depth calculations when gates are fully decomposed or nested in different configurations.

- **Tests**
  - Added regression coverage for recursive gate definitions, repeated gate calls, and nested external-gate depth scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Types of changes:
 ### Removed
 
 ### Fixed
+- Fixed an indirect cycle between gate definitions exhausting the Python stack: `gate a q { b q; }` with `gate b q { a q; }` raised a bare `RecursionError` naming nothing, while the direct case was already reported cleanly. The guard compared the body's gate name against one name, so it saw only a cycle of length one. It now tests membership of the whole expansion chain, and names the path: `Recursive definitions not allowed for gate 'a' (a -> b -> a)`. A gate reached twice down separate paths is a diamond, not a cycle, and still expands. ([#369](https://github.com/qBraid/pyqasm/issues/369))
 - Fixed a nested external custom gate counting the depth of the decomposition it skipped, the shape the [#352](https://github.com/qBraid/pyqasm/issues/352) fix did not reach: `unroll(external_gates=["outer"])` on a gate whose body calls another custom gate emitted one statement but reported `depth() == 13`. The suppression flag was assigned and cleared without save-restore, so the inner gate clobbered the outer gate's state in both directions. It is now saved and restored, and the depth is recorded once, from the outermost external gate. ([#367](https://github.com/qBraid/pyqasm/issues/367))
 
 ### Dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Types of changes:
 ### Removed
 
 ### Fixed
+- Fixed a nested external custom gate counting the depth of the decomposition it skipped, the shape the [#352](https://github.com/qBraid/pyqasm/issues/352) fix did not reach: `unroll(external_gates=["outer"])` on a gate whose body calls another custom gate emitted one statement but reported `depth() == 13`. The suppression flag was assigned and cleared without save-restore, so the inner gate clobbered the outer gate's state in both directions. It is now saved and restored, and the depth is recorded once, from the outermost external gate. ([#367](https://github.com/qBraid/pyqasm/issues/367))
 
 ### Dependencies
 
--- a/src/pyqasm/visitor.py
+++ b/src/pyqasm/visitor.py
@@ -1287,24 +1287,6 @@ class QasmVisitor:
             error_node=statement,
         )
 
-    def _raise_recursive_gate_error(self, gate_op: qasm3_ast.QuantumGate) -> None:
-        """Report a cyclic gate definition, naming the cycle it closes (issue #369).
-
-        Args:
-            gate_op (QuantumGate): The call in the gate body that closes the cycle.
-
-        Raises:
-            ValidationError: Always.
-        """
-        gate_name = gate_op.name.name
-        cycle = self._gate_expansion_chain[self._gate_expansion_chain.index(gate_name) :]
-        raise_qasm3_error(
-            f"Recursive definitions not allowed for gate '{gate_name}' "
-            f"({' -> '.join(cycle + [gate_name])})",
-            error_node=gate_op,
-            span=gate_op.span,
-        )
-
     def _visit_custom_gate_operation(
         self,
         operation: qasm3_ast.QuantumGate,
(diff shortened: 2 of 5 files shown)


TheGupta2012:
coderabbit generated no comments, merging

Issue #54, opened 2024-10-29
[BUG] Handle `opaque` in parsing qasm2 strings 


TheGupta2012:
### Environment

- **PyQASM version**: 0.0.2
- **Python version**: 3.11.6
- **Operating system**: macOS 14.4 

### What happened?

PyQASM should gracefully parse the `opaque` identifier - 

```python 
ValidationError           Traceback (most recent call last)
Cell In[6], line 1
----> 1 mod = pyqasm.load("""
      2 OPENQASM 2.0;
      3 include "qelib1.inc";
      4 qreg q[2];
      5 creg c[2];
      6 opaque custom_gate (a,b,c) p,q,r;""")

File ~/Desktop/qBraid/repos/pyqasm/pyqasm/entrypoint.py:47, in load(program)
     45         program = openqasm3.parse(program)
     46     except openqasm3.parser.QASM3ParsingError as err:
---> 47         raise ValidationError(f"Failed to parse OpenQASM string: {err}") from err
     48 elif not isinstance(program, openqasm3.ast.Program):
     49     raise TypeError("Input quantum program must be of type 'str' or 'openqasm3.ast.Program'.")

ValidationError: Failed to parse OpenQASM string:
```

### Suggestions (Optional)

Handle `opaque` in the visitor

ryanhill1:
Still reproducible on **pyqasm 1.0.4** (`main` @ 5dcae69), and we hit it in production-shaped work rather than synthetically, so adding evidence and a correction to the diagnosis.

### The suggestion in the description won't fix it

> Handle `opaque` in the visitor

The visitor never runs. `opaque` is **OpenQASM 2.0** syntax that was removed in OpenQASM 3, and pyqasm routes qasm2 through the `openqasm3` parser, so this dies at parse time:

```python
from openqasm3 import parse
parse('OPENQASM 2.0;\nopaque Rz(lam) q;\nqreg q[1];\n')   # QASM3ParsingError
parse('OPENQASM 2.0;\ngate mygate(a) q { rz(a) q; }\n')   # parses fine
```

The string `opaque` appears **nowhere** in the installed `openqasm3` package — there is no grammar production for it. So the fix has to live in pyqasm's own preprocessing (`preprocess.py` already rewrites source text to inline includes) or upstream in the grammar, not in the visitor.

### Real-world driver: Quantinuum's `hqslib1.inc`

[...]

Issue #370, opened 2026-08-15
opaque declarations fail to parse, blocking Quantinuum hqslib1 and any qasm2 file using them


ryanhill1:
### Environment

- **PyQASM version**: 1.0.4 (`main` @ 5dcae69)
- **Python version**: 3.11
- **Operating system**: macOS

### What happened?

pyqasm cannot parse any OpenQASM 2 program containing an `opaque` declaration. This blocks vendor include files — notably Quantinuum's `hqslib1.inc`, whose hardware primitives are all declared `opaque` — so no compiled program from Quantinuum H-series hardware can be loaded.

```python
import pyqasm
pyqasm.loads('''OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
opaque custom_gate(a,b,c) p,q,r;
''')
# ValidationError: Failed to parse OpenQASM string:
```

### Root cause: this fails at parse time, not in the visitor

`opaque` is OpenQASM **2.0** syntax that was removed in OpenQASM 3. pyqasm routes qasm2 through the `openqasm3` parser, which has no grammar production for it — the string `opaque` appears **nowhere** in the installed `openqasm3` package.

```python
from openqasm3 import parse
parse('OPENQASM 2.0;\nopaque Rz(lam) q;\nqreg q[1];\n')   # QASM3ParsingError
parse('OPENQASM 2.0;\ngate mygate(a) q { rz(a) q; }\n')   # parses fine
```

So the failure happens before any pyqasm visitor code runs, and a fix in the visitor cannot reach it. It has to live either in pyqasm's own source preprocessing (`preprocess.py` already rewrites program text to inline includes, so it is well placed for this) or upstream in the grammar.

### Real-world driver: Quantinuum `hqslib1.inc`

`hqslib1.inc` (shipped in `pytket/qasm/includes/`) is Quantinuum's qasm2 include file, the vendor counterpart to `qelib1.inc`. It opens with six `opaque` declarations — these are the machine primitives, so they intentionally have no decomposition:

```qasm
opaque Rz(lam) q;
opaque U1q(theta, phi) q;
opaque ZZ() q1,q2;
opaque RZZ(theta) q1,q2;
opaque Rxxyyzz(alpha, beta, gamma) q1,q2;
opaque Rxxyyzz_zphase(alpha, beta, gamma, z0, z1) q1,q2;
```

A bell state compiled for H-series hardware looks like the following, and currently cannot be loaded at all:

```qasm
OPENQASM 2.0;
[...]

PR #376, opened 2026-08-18
fix: detect gate definition cycles of any length, not just self-calls

### LINKED ISSUE #369: Indirect gate recursion (a -> b -> a) raises RecursionError instead of ValidationError

TheGupta2012:
Fixes #369.

## The bug

pyqasm reported a gate that calls itself cleanly, but a cycle between two definitions exhausted the Python stack:

```python
# direct — already correct
pyqasm.loads('gate a(t) q { a(t) q; } ...').validate()
# ValidationError: Recursive definitions not allowed for gate 'a'

# indirect — stack blowout
pyqasm.loads('gate a(t) q { b(t) q; } gate b(t) q { a(t) q; } ...').validate()
# RecursionError: maximum recursion depth exceeded
```

The `RecursionError` named nothing, so there was no indication of which gates were cycling.

## Cause

The guard in `_visit_custom_gate_operation` compared the body's gate name against **one** name — the gate currently being expanded:

```python
if isinstance(gate_op, qasm3_ast.QuantumGate) and gate_op.name.name == gate_name:
```

A cycle of length one matches. A cycle of length two or more passes the check and recurses through gate expansion until the interpreter's stack limit.

## The fix

Track the chain of gates currently being expanded (`_gate_expansion_chain`, pushed on entry and popped in a `finally`) and test **membership of that chain** instead of equality with a single name. One mechanism now covers both cases, so the direct-recursion check is replaced rather than duplicated.

The error names the path it closes, as the issue requested:

```
Recursive definitions not allowed for gate 'a' (a -> b -> a)
Recursive definitions not allowed for gate 'a' (a -> b -> c -> a)
Recursive definitions not allowed for gate 'a' (a -> a)          # direct, unchanged shape
```

The check stays at the same point in the body scan as the old one, so the reported span is still the call inside the definition that closes the cycle.

**Chain, not a seen-set.** Membership of the *chain* is what distinguishes a cycle from a diamond. With a set of every gate seen, `a` calling both `b` and `c`, where both call `d`, would be misreported as recursion. Popping on return keeps that legal.

## Behaviour that does not change

| Program | Before | After |
|---|---|---|
| `gate a q { a q; }` | `ValidationError` | same, plus `(a -> a)` |
| gate `b` never defined | `Unsupported / undeclared QASM operation: b` | unchanged |
| diamond `a → {b, c} → d` | expands | expands |
| same gate called twice in one body | expands twice | expands twice |

## Verification

- Full suite: **789 passed, 4 skipped**.
- **Mutation-tested.** With the source fix reverted, all three new cycle tests fail with `RecursionError`; the two diamond/repeat guards pass on `main`, as they should.

## Tests added

- `CUSTOM_GATE_INCORRECT_TESTS` gains `indirect_recursive_definition` (a → b → a) and `three_gate_recursive_definition` (a → b → c → a). Both pin the full message including the path, and both inherit the existing line/column assertions.
- `test_shared_gate_definition_is_not_recursion` — the diamond must expand, not raise.
- `test_repeated_gate_call_in_one_body_is_not_recursion` — the chain entry must be popped when an expansion returns.
- `test_indirect_recursion_does_not_exhaust_the_stack` — pins that the failure is a `ValidationError` and never a `RecursionError`, which was the reported symptom.

## Note on the base branch

This is stacked on #375, because both PRs edit `_visit_custom_gate_operation` — reviewing them independently would hand you a conflict. **Merge #375 first**; this diff then reduces to its own four files. The commits are separate and each stands on its own.

## Follow-up not taken

The issue notes this is likely to be hit by anyone fixing #54 / #370 (`opaque` declarations), since giving Quantinuum's opaque primitives real bodies creates a mutual cycle with `hqslib1`'s `U` and `CX`. That case now reports the cycle by name instead of blowing the stack, which is the whole of what this issue asked for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Types of changes:
 ### Removed
 
 ### Fixed
+- Fixed an indirect cycle between gate definitions exhausting the Python stack: `gate a q { b q; }` with `gate b q { a q; }` raised a bare `RecursionError` naming nothing, while the direct case was already reported cleanly. The guard compared the body's gate name against one name, so it saw only a cycle of length one. It now tests membership of the whole expansion chain, and names the path: `Recursive definitions not allowed for gate 'a' (a -> b -> a)`. A gate reached twice down separate paths is a diamond, not a cycle, and still expands. ([#369](https://github.com/qBraid/pyqasm/issues/369))
 - Fixed a nested external custom gate counting the depth of the decomposition it skipped, the shape the [#352](https://github.com/qBraid/pyqasm/issues/352) fix did not reach: `unroll(external_gates=["outer"])` on a gate whose body calls another custom gate emitted one statement but reported `depth() == 13`. The suppression flag was assigned and cleared without save-restore, so the inner gate clobbered the outer gate's state in both directions. It is now saved and restored, and the depth is recorded once, from the outermost external gate. ([#367](https://github.com/qBraid/pyqasm/issues/367))
 
 ### Dependencies
--- a/src/pyqasm/visitor.py
+++ b/src/pyqasm/visitor.py
@@ -124,6 +124,9 @@ class QasmVisitor:
         self._function_qreg_transform_map: deque = deque([])  # for nested functions
         self._global_creg_size_map: dict[str, int] = {}
         self._custom_gates: dict[str, qasm3_ast.QuantumGateDefinition] = {}
+        # gates currently being expanded, outermost first; a name reappearing on it is a
+        # cycle of any length between gate definitions (issue #369)
+        self._gate_expansion_chain: list[str] = []
         self._external_gates: list[str] = [] if external_gates is None else external_gates
         self._subroutine_defns: dict[
             str, qasm3_ast.SubroutineDefinition | qasm3_ast.ExternDeclaration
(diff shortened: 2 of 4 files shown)


TheGupta2012:
Looks good, the error was simple

Issue #368, opened 2026-08-15
loads() does not resolve custom include statements, unlike load()


ryanhill1:
### Environment

- **PyQASM version**: 1.0.4 (`main` @ 5dcae69)
- **Python version**: 3.11
- **Operating system**: macOS

### What happened?

`load()` resolves custom `include` statements; `loads()` silently does not. The two entry points are otherwise documented as equivalent ("Loads an OpenQASM program into a `QasmModule` object"), so the asymmetry is easy to hit and the failure is misleading when you do.

`entrypoint.py`:

```python
def load(filename: str, **kwargs) -> QasmModule:
    ...
    program = process_include_statements(filename)   # <-- only here
    return loads(program, **kwargs)


def loads(program: openqasm3.ast.Program | str, **kwargs) -> QasmModule:
    ...                                              # no include resolution
```

`process_include_statements` takes a filename and resolves includes relative to the file's own path (`_resolve_include_path`), so there is nothing for the string path to key off.

### Reproduction

Given a program that includes a custom file:

```qasm
OPENQASM 2.0;
include "mygates.inc";
qreg q[2];
mygate q[0];
```

- `pyqasm.load("prog.qasm")` with `mygates.inc` alongside it → include inlined correctly
- `pyqasm.loads(open("prog.qasm").read())` → include ignored, then fails later with

```
ValidationError: Unsupported / undeclared QASM operation: mygate
```

The error names the gate, not the unresolved include, which sends you looking at the gate table rather than at include handling. It took reading `preprocess.py` to work out that the string path never attempts resolution at all.

### Why it matters

Programs frequently arrive as strings rather than files — fetched from object storage, returned by a vendor API, or passed between services — with no filesystem location to resolve relative includes against. In our case a compiled program is retrieved from cloud storage and handed to pyqasm as a string; there is no file, so `load()` is not an option and includes can never resolve.

### Suggestions (Optional)

`loads()` already accepts injected definitions via the `extern_functions` kwarg, which suggests a natural shape here — something like:

```python
pyqasm.loads(src, include_sources={"mygates.inc": "<file contents>"})
# or
pyqasm.loads(src, include_path=["/path/to/includes"])
```

Failing that, raising an error that names the unresolved include (rather than the downstream undeclared gate) would remove most of the confusion.

Related: #54 — `opaque` parsing. Both must be addressed for vendor include files such as Quantinuum's `hqslib1.inc` to be usable from a string.

PR #378, opened 2026-08-18
feat: support OpenQASM 2 opaque declarations as black-box gates

### LINKED ISSUE #370: opaque declarations fail to parse, blocking Quantinuum hqslib1 and any qasm2 file using them

TheGupta2012:
Fixes #370. Supersedes #54.

## The problem

pyqasm could not parse **any** OpenQASM 2 program containing an `opaque` declaration:

```python
pyqasm.loads('OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[2];\nopaque custom_gate(a,b,c) p,q,r;\n')
# ValidationError: Failed to parse OpenQASM string:
```

This blocked vendor include files — notably Quantinuum's `hqslib1.inc`, whose six hardware primitives are all declared `opaque` — so no compiled program from H-series hardware could be loaded at all.

**The failure is at parse time, not in the visitor.** `opaque` is OpenQASM 2 syntax that OpenQASM 3 removed, and pyqasm routes qasm2 through the `openqasm3` parser, which has no production for it — the string `opaque` appears nowhere in the installed `openqasm3` package. This is why #54's suggestion to handle it in the visitor could not work.

## The change

The rewrite lives in source preprocessing, where include inlining already happens:

```
opaque Rz(lam) q;   ->   gate Rz(lam) q { }
```

plus the name recorded on the module in `_opaque_gates`. An opaque gate has no decomposition by definition, so the empty body is a carrier for its **name and arity** only; the recorded name is what makes it a black box.

- **Routing.** `_visit_generic_gate_operation` sends an opaque call to `_visit_external_gate_operation`, so it is emitted as written and counts as one layer of depth — exactly the treatment an external gate gets. A new `_is_black_box_gate()` helper collects the three cases that share this treatment (external, verbatim-box, opaque).
- **Never flushed.** `unroll()` resets `external_gates` on every call. Opaque gates are stored on the module instead, because they are a property of the program and have no decomposition to fall back on. Parametrized test over `None` / `[]` / `["h"]` / `["ZZ"]`.
- **Gated on the version.** The rewrite only runs when the `OPENQASM 2` header matches, so `opaque` in a qasm3 program keeps failing to parse.
- **Runs after include inlining**, so an `opaque` inside a vendor include file is reached by both `load()` and `loads(..., include_dir=...)`.

## What works

The issue's bell state, with an `hqslib1`-shaped include, loads and unrolls:

```
qreg q[2]; creg c[2];
rz(3.141592653589793) q[0];
U1q(1.5707963267948966, 1.5707963267948966) q[0];
RZZ(1.5707963267948966) q[0], q[1];
measure q[0] -> c[0];
```

`validate()`, `depth()`, `has_measurements()`, `num_qubits`, `remove_idle_qubits()`, `reverse_qubit_order()` and `unroll(consolidate_qubits=True)` all handle opaque calls, and arity is validated (`ZZ q[0];` → `Qubit count mismatch for gate 'ZZ'. Expected 2 qubits, but got 1`).

Better than the issue predicted: it also handles `hqslib1`'s `U` and `CX`, which are defined *in terms of* the opaque primitives. The issue author had to drop those; they now unroll down to the primitives and stop there.

## Three limits, all deliberate, all tested

| Behaviour | Why |
|---|---|
| The declaration is **not re-emitted** in the unrolled output | Matches the documented precedent for external gates, whose `gate` definition unrolling already drops. Consequence: the output has calls to a gate it does not declare, so it does not load back into pyqasm on its own. A printer hook for re-emitting `opaque` is deliberately not built in v1. |
| `rebase()` raises `RebaseError: Gate 'ZZ' is not supported in the '<basis>' set` | An opaque primitive has no decomposition, so it cannot be rewritten onto a standard basis set. It reaches the existing unsupported-gate path and is named there — not a crash, and no change was needed. |
| **`to_qasm3()` now raises** for a program declaring one | A body-less `gate` means the *identity* in OpenQASM 3. Emitting the carrier definition would silently turn every hardware primitive into a no-op. OpenQASM 3 removed `opaque` and has no equivalent, so there is no correct translation and refusing is the loud option. Nothing inside `src/` calls `to_qasm3`, so this is confined to the public API, and these programs could not be loaded before this PR — no workflow regresses. Easily relaxed if someone proposes a representation. |

All three are documented in a new **Opaque gates** section in `src/README.md`, alongside the existing Pragmas notes.

## Verification

- Full suite: **817 passed, 4 skipped**. No existing expectation changed.
- **Mutation-tested.** With the four source files reverted, **19 of 20** opaque tests fail. The one that passes is `test_opaque_is_not_qasm3_syntax` — correctly, since it pins behaviour this PR must *not* change.
- The `hqslib1`-like fixture is written out in the test file rather than vendored, so nothing depends on pytket being installed.

## Merge order

This is stacked on **#377** (`include_dir`), because both touch `loads()` and `preprocess.py`. Merge #377 first.

**The #375 conflict is already resolved.** #375 (issues #367 and #369) has merged, and this branch is rebased on top of it. Both PRs rewrote the `_recording_ext_gate_depth` assignment at the top of `_visit_custom_gate_operation`; the resolution keeps #375's save/restore shape with this PR's helper:

```python
prev_recording = self._recording_ext_gate_depth
is_external = self._is_black_box_gate(gate_name)      # <- not the inlined condition
self._recording_ext_gate_depth = prev_recording or is_external
```

Resolving it the other way — keeping `self._in_verbatim_box or gate_name in self._external_gates` — would make opaque gates record **zero** depth, because `is_external` also feeds the single depth-recording block below. `test_opaque_gate_counts_as_one_towards_depth` covers it.

## Related

- **#368 / #377** — the other half of loading a Quantinuum program from a string: #377 lets `hqslib1.inc` be supplied as a string, this PR lets its `opaque` lines parse. Both are needed; each is independent.
- **#369 / #376** — the issue predicted whoever fixed this would hit indirect gate recursion, since giving the opaque primitives real bodies creates a mutual cycle with `hqslib1`'s `U` and `CX`. That is now reported by name instead of blowing the stack. This PR removes the need for that workaround entirely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Types of changes:
 ## Unreleased
 
 ### Added
+- Added support for OpenQASM 2 `opaque` declarations, which previously failed at parse time and blocked vendor include files such as Quantinuum's `hqslib1.inc`. An opaque gate is treated as a black box: emitted as written, counted as one layer of depth. `to_qasm3()` rejects such a program. ([#370](https://github.com/qBraid/pyqasm/issues/370))
 - Added an `include_dir` kwarg to `loads()` and `load()`, naming the directory custom `include` statements resolve against. A program given as a string could not resolve includes at all, and failed later naming the gate rather than the include. Resolution is opt-in: without the kwarg, no files are read. ([#368](https://github.com/qBraid/pyqasm/issues/368))
 
 ### Improved / Modified
--- a/src/README.md
+++ b/src/README.md
@@ -41,6 +41,35 @@ Source code for OpenQASM 3 program validator and semantic analyzer
 | ComplexType                    | ✅          | Completed              |
 | AngleType                      | ✅          | Completed              |
 | ExternDeclaration              | ✅          | Completed              |
+| opaque (OpenQASM 2 only)       | ✅          | Emitted as written     |
+
+## Opaque gates
+
+`opaque NAME(params) qubits;` is OpenQASM 2 syntax that was removed in OpenQASM 3. It
+declares a hardware primitive: a gate with a name and an arity and no decomposition.
+Quantinuum's `hqslib1.inc` opens with six of them.
+
+pyqasm rewrites the declaration before parsing and records the name, so `validate()`,
+`depth()`, `has_measurements()` and the qubit-renumbering passes all work. A call to an
+opaque gate is emitted as written rather than unrolled, and counts as one layer of depth,
+exactly as an external gate does. `unroll()` never drops that treatment: unlike
+`external_gates`, which the caller sets per call, an opaque gate is a property of the
+program and has no decomposition to fall back on.
+
+Two things to know:
+
+- **The declaration is not re-emitted.** Unrolling drops it, the same way it drops the
+  `gate` definition of an external gate, so the output carries calls to a gate it does
+  not declare and does not load back into pyqasm on its own.
+- **`opaque` in an OpenQASM 3 program is still a parse error.** The rewrite is gated on
+  the `OPENQASM 2` header, because the keyword is not OpenQASM 3 syntax.
+- **`rebase()` reports it as unsupported.** An opaque primitive has no decomposition, so
+  it cannot be rewritten onto a standard basis set; it reaches the existing
+  unsupported-gate path and is named there.
+- **`to_qasm3()` refuses a program that declares one.** OpenQASM 3 removed `opaque` and
+  has no equivalent for a gate with no decomposition, and a body-less `gate` in
+  OpenQASM 3 means the identity — so converting would silently turn each hardware
+  primitive into a no-op.
 
 ## Pragmas
 
--- a/src/pyqasm/entrypoint.py
+++ b/src/pyqasm/entrypoint.py
@@ -27,7 +27,11 @@ import openqasm3
 from pyqasm.exceptions import ValidationError
 from pyqasm.maps import SUPPORTED_QASM_VERSIONS
 from pyqasm.modules import Qasm2Module, Qasm3Module, QasmModule
-from pyqasm.preprocess import process_include_sources, process_include_statements
+from pyqasm.preprocess import (
+    process_include_sources,
+    process_include_statements,
+    rewrite_opaque_declarations,
+)
 
 if TYPE_CHECKING:
     import openqasm3.ast
@@ -165,9 +169,12 @@ def loads(program: openqasm3.ast.Program | str, **kwargs) -> QasmModule:
     """
     _validate_kwargs(kwargs)
     include_dir = kwargs.pop("include_dir", None)
+    opaque_gates: set[str] = set()
     if isinstance(program, str):
         if include_dir is not None:
             program = process_include_sources(program, include_dir)
+        # after include resolution, so an opaque in a vendor include is rewritten too
+        program, opaque_gates = rewrite_opaque_declarations(program)
         try:
             program = openqasm3.parse(program)
         except openqasm3.parser.QASM3ParsingError as err:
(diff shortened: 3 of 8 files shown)


ryanhill1 (dismissed):

ryanhill1 (dismissed):

TheGupta2012:
Merging as it was previously reviewed by Ryan, just rebasing dismissed the review.

PR #405, opened 2026-08-24
Keep pyqasm validate diagnostics on one line

TheGupta2012:
## Problem

`pyqasm validate` builds its output with a default `rich.Console`, which wraps at the detected console width (80 when stdout is not a tty). A file path longer than that is broken **mid-token** across lines:

```
/private/tmp/.../3eba83a3-89e
0-4a99-a5cf-386bf6fb273b/scratchpad/...: error: ...
```

The reported path is then neither copyable nor clickable in a terminal.

The same wrapping breaks two CLI tests, which assert the absolute path appears in the output:

- `tests/cli/test_cli_commands.py::test_validate_qasm_with_invalid_file`
- `tests/cli/test_cli_commands.py::test_validate_command_with_invalid_file`

They pass in CI only because the runner checks out to a short path (`/home/runner/work/pyqasm/pyqasm`). They fail for any contributor whose clone sits at a longer path, which is why they showed up as "pre-existing failures" while reviewing #402, #403 and #404.

## Fix

The error console now uses `soft_wrap=True`, which disables wrapping and cropping while leaving markup intact — one diagnostic per line, path unbroken.

## Verification

`tests/cli` at a short path and at a path long enough to trigger the old wrap:

| | before | after |
|---|---|---|
| short path | 33 passed | 33 passed |
| long path | 2 failed, 31 passed | 33 passed |

Full suite at the long path: **802 passed, 3 skipped**. `pylint` 10.00/10, `black` and `isort` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `pyqasm validate` diagnostics so long file paths remain intact instead of splitting across terminal lines.
- **Documentation**
  - Added an unreleased changelog entry describing the validation output improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Types of changes:
 ### Removed
 
 ### Fixed
+- Fixed `pyqasm validate` wrapping its diagnostics at the console width, which split a file path longer than the width across lines mid-token and left it neither copyable nor clickable. The error console now uses `soft_wrap`, keeping one diagnostic per line.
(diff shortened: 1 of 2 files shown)


TheGupta2012:
Merging as the `coderabbit` review has come through without changes

PR #411, opened 2026-08-24
ci: publish to PyPI via trusted publishing instead of an API token

ryanhill1:
Matches qBraid/qBraid#1346, which does the same for the SDK.

The publish workflows authenticated to PyPI with `PYPI_API_TOKEN`, a long-lived
credential stored as a repository secret. This switches them to Trusted Publishing:
the job requests `id-token: write`, and `pypa/gh-action-pypi-publish` mints a
short-lived OIDC credential scoped to that one workflow. No publishing secret is
stored in the repository afterwards.

Trusted Publishing is also the prerequisite for PEP 740 attestations, which the
action generates by default. Those attach the repository, workflow and commit SHA
to each uploaded file, so a published wheel can be traced back to the commit that
built it. They apply to releases published after this merges, not retroactively.

Trusted publishing has been configured on PyPI for both `release.yml` and `pre-release.yml`


TheGupta2012 (dismissed):

TheGupta2012:
Seems like the [publish step did not have the right permissions](https://github.com/qBraid/pyqasm/actions/runs/32815004031/job/97702314457), I think you accidentally added them in [publish sdist step](https://github.com/qBraid/pyqasm/pull/411/changes/d8bc0e6503386749230df25442b0d6e026d6ea94#diff-8f0c4d810b263ab478547020c847ee0297cc98c11c67fc2209ce604e52719607R126).

I've added them in the `publish-to-pypi` flow but that led me to a bug in our pre-release workflow! Seems like we weren't updating the `sdist` names correctly for pre-releases. Raised the PR #413 for that, should fix it

TheGupta2012:
Completed pre-release - https://github.com/qBraid/pyqasm/actions/runs/32821887958/job/97722479493

Had to bump to `1.2.0-alpha` for the new pre-release

TheGupta2012 (approved):

PR #412, opened 2026-08-24
ci: add OpenSSF Scorecard workflow

ryanhill1:
Adds the OpenSSF Scorecard workflow, copied verbatim from `qbraid-algorithms`, where it
is already running and publishing a score to the public Scorecard API.

Scorecard grades a repository against supply-chain practices — branch protection, pinned
dependencies, token permissions, publishing provenance, code review — and publishes the
result at `api.securityscorecards.dev`, where anyone can look it up. The value is that a
third party computes the number rather than us.

Notes on the workflow itself, carried over from the original:

- Action versions are pinned to full commit SHAs rather than tags, because a tag can be
  repointed at other code, and that is exactly what Scorecard's own Pinned-Dependencies
  check looks for.
- Top-level `permissions: {}` means a job added later starts with nothing until it
  declares what it needs.
- `concurrency` cancels superseded runs, since Scorecard reports current repository
  state and an outdated run has nothing to add.

Results upload to code scanning as SARIF and to the public API. No secrets are required;
`id-token: write` is what lets the run publish under the repository's identity.


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,61 @@
...
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write   # upload SARIF to code scanning
+      id-token: write          # publish results to the public Scorecard API
+      contents: read
+      # Lets Scorecard read workflow definitions and run history. Documented by the
+      # upstream starter workflow as required for private repositories; this repo is
+      # public, so it is retained only as a safeguard should that ever change.
+      actions: read
```
TheGupta2012: Is this needed? As per the comment, it is present only as a safeguard if we choose to make `pyqasm` private , which might not be happening.

TheGupta2012 (commented):
@ryanhill1 looks good, just one comment regarding permissions

TheGupta2012:
Haven't tested it 'cause we'll need it to be merged to `main` for it to appear in the actions.

.github/workflows/scorecard.yml
ryanhill1: You're right, and it goes one line further than you flagged — `contents: read` has exactly the same status. Both removed in dd81908.

I dug through the OpenSSF sources rather than reason from the comment, since my own justification for the line turned out to be the weak part. Evidence, strongest last:

**1. The endpoint Scorecard needs is publicly readable.** Its only Actions API call is `ListWorkflowRunsByFileName` (`checks/raw/github/packaging.go`, `checks/raw/binary_artifact.go`), hitting `GET /repos/{owner}/{repo}/actions/workflows/{file}/runs`. Requesting that with **no authentication at all** returns `200`; `/user` returned `401` in the same session, so the request really was anonymous. A token cannot have less access than anonymous on a public repo.

**2. Upstream ships both commented out.** `actions/starter-workflows` `code-scanning/scorecard.yml`:
```yaml
      # Uncomment the permissions below if installing in a private repository.
      # contents: read
      # actions: read
```
So my comment's premise was right — private-repo only — and the conclusion I drew from it was wrong. Keeping it "as a safeguard" just meant carrying a permission that does nothing, which is what a least-privilege review should strip.

**3. OSSF dogfoods it.** `ossf/scorecard` and `ossf/scorecard-action` are both public and run with `security-events: write` and `id-token: write` alone. Last five runs green, and the checks that touch the Actions API score **Packaging 10, Binary-Artifacts 10, CI-Tests 10**. That is also what justifies dropping `contents: read` — their `actions/checkout` works without it.

**Residual risk, stated plainly:** both affected checks return an error rather than degrading if the call fails, so a mistake here surfaces as a check erroring, not as a silently lower score. The same workflow has already run green on qBraid and qbraid-algorithms with the wider permissions, giving a baseline to compare the next run against.

Same change: qBraid/qBraid#1354 and qBraid/qbraid-algorithms#97 (both merged their scorecard PRs before this came up, so they needed follow-ups).

TheGupta2012 (approved):

PR #419, opened 2026-08-26
ci: cancel superseded runs and trim the PR wheel matrix

kanavsetia:
## Why

`pyqasm` is the largest single contributor to the org's GitHub Actions bill. In the August billing period it accounted for **~$219 of ~$407 gross Actions spend** — and almost all of it is macOS:

```
$196.29   3,166 min @ $0.062/min   Actions macOS 3-core    (~10x the Linux rate)
$ 22.65   2,265 min @ $0.010/min   Actions Windows
```

The effect shows up clearly in the org-wide effective rate — August used **36% fewer** Actions minutes than July but cost **more**, because the per-minute rate nearly doubled:

| Month | Minutes | Gross | $/min |
|---|---:|---:|---:|
| June | 38,530 | $247.52 | $0.00642 |
| July | 55,083 | $371.60 | $0.00675 |
| Aug (MTD) | 34,998 | $407.32 | **$0.01164** |

macOS minutes went from **658 in July to 3,309 in August**.

Two causes:

1. **No concurrency guard.** `synchronize` fires on every push to an open PR, so each push started another full 20-entry wheel matrix while the previous one kept running to completion.
2. **Every PR built the complete matrix** — 5 Python versions × Windows / Linux / macOS x86_64 / macOS arm64, i.e. **10 macOS jobs per run** — even for a one-line docs change.

## What this changes

**1. Concurrency guard**

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Superseded PR runs are cancelled. Pushes to `main` are never cancelled.

**2. Per-event matrix selection**

| Event | Entries | Composition |
|---|---:|---|
| `pull_request` | **7** | Linux × 3.10–3.14, macOS arm64 × 3.11, Windows × 3.11 |
| `push` to main / `workflow_dispatch` | **20** | unchanged — the full original matrix |

macOS jobs per PR run drop from **10 → 1**.

## Invariants preserved

- **The full 20-entry matrix is byte-for-byte the same set as the previous inline list** (verified by diffing the parsed entries), so `main` and manual runs behave exactly as before.
- **macOS arm64 / 3.11 is retained in the PR set on purpose** — the `code-coverage` job consumes the artifact that specific entry uploads (`if: matrix.os == 'macos-14' && matrix.python == 311`). Dropping it would silently break coverage.
- **Published wheels are unaffected.** PyPI wheels are built by `release.yml` and `pre-release.yml`, which each carry their own complete matrix. `main.yml` is CI validation only.

## Tradeoff

PRs no longer validate wheel builds on Windows 3.10/3.12/3.13/3.14, macOS x86_64 at all, or macOS arm64 outside 3.11. A platform-specific build break in those combinations would surface on merge to `main` rather than on the PR — and still before any release, since `release.yml` builds the full matrix independently.

If that feels too loose, the natural middle ground is adding `macos-14 / 3.14` to the PR set to cover the newest Python on arm64; that's one extra macOS job instead of nine.


.github/workflows/main.yml
```diff
@@ -7,92 +7,98 @@ on:
...
+          # the code-coverage job consumes the artifact this entry uploads;
+          # Windows 3.11 is kept as a smoke check for the non-POSIX path.
+          read -r -d '' PR_MATRIX <<'JSON' || true
+          {"include":[
+            {"os":"ubuntu-latest","python":310,"platform_id":"manylinux_x86_64","manylinux_image":"manylinux_2_28"},
+            {"os":"ubuntu-latest","python":311,"platform_id":"manylinux_x86_64","manylinux_image":"manylinux_2_28"},
+            {"os":"ubuntu-latest","python":312,"platform_id":"manylinux_x86_64","manylinux_image":"manylinux_2_28"},
+            {"os":"ubuntu-latest","python":313,"platform_id":"manylinux_x86_64","manylinux_image":"manylinux_2_28"},
+            {"os":"ubuntu-latest","python":314,"platform_id":"manylinux_x86_64","manylinux_image":"manylinux_2_28"},
+            {"os":"macos-14","python":311,"platform_id":"macosx_arm64"},
```
TheGupta2012: We should at least keep the `macos-15-intel` image in the tests. Currently, `x86` arch is not being tested for MacOS.

TheGupta2012:
Thanks for catching this @kanavsetia . However, embedding the choice matrix in a string might not be the best path from a maintenance standpoint. Can you figure out if there's a better approach to cleanly switch the python and os matrix based on whether we are running in a PR or from `main`?

TheGupta2012 (dismissed):

TheGupta2012 (approved):

PR #424, opened 2026-09-10
Support OpenQASM end statements

### LINKED ISSUE #396: `end;` statement is unsupported
## Limitation

The [classical instructions](https://openqasm.com/versions/3.1/language/classical.html) chapter defines: "The statement
`end;` immediately terminates the program, no matter what scope it appears in."

`pyqasm` has no handler for `openqasm3.ast.EndStatement` and falls through to the generic
unsupported-statement error.

## Example QASM failure

```qasm
OPENQASM 3.0;
include "stdgates.inc";
qubit[1] q;
h q[0];
end;
```

```
ValidationError: Unsupported statement of type <class 'openqasm3.ast.EndStatement'>
```

Inside a branch, which is the form that carries real meaning:

```qasm
OPENQASM 3.0;
int[8] i = 0;
if (i == 0) { end; }
// ValidationError: Unsupported statement of type <class 'openqasm3.ast.EndStatement'>
```

## Change Requested

1. Accept `end;` wherever a statement is allowed — global scope, loop bodies, `if`/`else`
   blocks, `switch` cases, `box` bodies, and subroutine bodies.
2. At global scope, stop visiting subsequent statements: everything after `end;` is
   unreachable and must not appear in the unrolled output.
3. Preserve `end;` through `dumps()`.
4. Inside a conditional block whose condition depends on a runtime value, `end;` cannot be
   resolved at analysis time. Emit it into the unrolled output as-is rather than attempting to
   truncate the program.

## Implementation Details

- Add an `EndStatement` branch to the statement dispatch in `src/pyqasm/visitor.py`; the
  current fallback that raises is at `visitor.py:3492`.
- The unconditional global-scope case is the simple one: set a flag and stop consuming
  statements at that level.
- Consider whether `QasmModule` should expose that the program contains a terminating `end;` —
  it is the kind of fact the existing `has_measurements()` / `has_barriers()` helpers report,
  and downstream consumers may want it.
- Sanity-check the depth and qubit-count bookkeeping: statements after a global `end;` must
  not contribute, since they are unreachable.
- Tests: `tests/qasm3/test_statements.py` — `end;` at global scope with trailing statements
  that must be dropped, `end;` inside `if`, inside a `for` body, inside a subroutine, and a
  `dumps()` round-trip.

danielgaskins:
## Summary

- accept OpenQASM 3 `end;` statements during validation and unrolling
- stop statically unrolled blocks after an unconditional terminator
- preserve conditional terminators inside runtime-dependent branches
- cover global, branch, loop, switch, box, and subroutine behavior

## Testing

- `pytest -q tests/qasm3/test_end.py`
- `pytest -q tests`
- `pytest -q --cov=pyqasm --cov-config=pyproject.toml --cov-report=term --cov-report=xml tests`
- `pylint src tests bin examples`
- `isort --check-only src tests bin examples`
- `black --check --target-version py310 src tests bin examples`
- `mypy src examples`
- `qbraid admin headers src tests bin examples --skip=src/pyqasm/_version.py --type=apache`
- `sphinx-build -W -b html docs/ <output-directory>`

Fixes #396


tests/qasm3/test_end.py
```diff
@@ -0,0 +1,211 @@
...
+        include "stdgates.inc";
+        qubit q;
+        {control_flow}
+        x q;
+        """)
+
+    module.unroll()
+    output = dumps(module)
+
+    assert output.count("h q[0];") == 1
```
TheGupta2012, line 76: Can you use already built test validators in our utilities? It seems like we already have a pretty extensive utility module built for these assertions.

src/pyqasm/visitor.py
```diff
@@ -3505,6 +3520,41 @@ def _visit_include(self, include: qasm3_ast.Include) -> list[qasm3_ast.Statement
...
+        Returns:
+            list[EndStatement]: The statement in a list, or an empty list if
+                self._check_only is True.
+        """
+        if self._check_only:
+            return []
+        return [statement]
+
+    @classmethod
+    def _ends_with_end_statement(cls, statements: Sequence[qasm3_ast.Statement]) -> bool:
```
TheGupta2012, line 3540: Why do you need all statements in this method as a parameter? I think we can just isolate the processing inside the `analyser.py` and refactor

TheGupta2012 (changes requested):
Hi @danielgaskins , thanks for extending `pyqasm` with this contribution! I just gave a couple of comments but otherwise looks good. Can you also add a `CHANGELOG` entry besides these changes?

danielgaskins:
Thanks, that makes sense. I moved the termination check into `Qasm3Analyzer` and now pass it the last emitted statement. The end tests use `check_unrolled_qasm`, and I added a changelog entry for #396. I updated the branch from main too. The full suite passes with 869 tests and 3 skips, and lint and typing checks are clean.

TheGupta2012 (approved):
Thanks @danielgaskins for the changes, they lgtm!

PR #425, opened 2026-09-10
Handle general expression statements

### LINKED ISSUE #388: Bare expression statements crash with `AttributeError` instead of raising `ValidationError`
## Limitation

An expression followed by a semicolon is a syntactically valid OpenQASM 3 statement — the
[classical types](https://openqasm.com/versions/3.1/language/classical.html) chapter uses this form throughout to
illustrate operator results, for example under
[Integers](https://openqasm.com/versions/3.1/language/classical.html#integers) and
[Angles](https://openqasm.com/versions/3.1/language/classical.html#angles).

`pyqasm` reaches these statements and dereferences `.name` on the expression node. For any
non-`Identifier` expression this raises an `AttributeError` that escapes the public API. The
same dereference fails for an undeclared gate applied to zero qubits.

## Example QASM failure

Straight from the spec's Integers section:

```python
import pyqasm

m = pyqasm.loads('''
OPENQASM 3.0;
int[32] a = 2;
int[32] b = 3;

a * b;   // 6
b / a;   // 1
b % a;   // 1
''')
m.validate()
```

```
AttributeError: 'BinaryExpression' object has no attribute 'name'
```

The spec's Angles section fails identically:

```qasm
OPENQASM 3.0;
angle[4] a = 7 * (pi / 8);
angle[4] b = pi / 8;
a + b;
// AttributeError: 'BinaryExpression' object has no attribute 'name'
```

And the gate-application example in
[Applying gates](https://openqasm.com/versions/3.1/language/gates.html#applying-gates):

```qasm
OPENQASM 3.0;
g1 q1;
// AttributeError: 'str' object has no attribute 'name'
```

## Change Requested

1. Never surface an `AttributeError` from a statement `pyqasm` does not handle. Every
   unhandled statement or expression form must raise `ValidationError` carrying the source
   span.
2. Decide and document the semantics of a bare expression statement. It has no effect on the
   quantum program, so the reasonable options are to evaluate and discard it, or to reject it
   with a clear message such as `"Expression statement has no effect"`. Either is acceptable;
   the current crash is not.
3. An undefined gate identifier must report the undefined name, not fail on an attribute
   lookup.

## Implementation Details

- The root cause is the visitor's dispatch table in `src/pyqasm/visitor.py:193`:

  ```python
  qasm3_ast.ExpressionStatement: lambda x: self._visit_function_call(x.expression),
  ```

  Every expression statement is assumed to be a function call, so `_visit_function_call`
  dereferences `x.expression.name` — which exists on an `Identifier` but not on a
  `BinaryExpression`, a `UnaryExpression`, an `IndexExpression`, or a literal.
- The fix is to branch on the expression type before dispatching: route `FunctionCall` to
  `_visit_function_call` and give every other form the chosen behaviour from point 2 above.
  The generic fallback at `visitor.py:3492` already raises the correct error shape and can be
  reused.
- The `'str' object has no attribute 'name'` variant comes from the gate-operation path, where
  the operation name has already been reduced to a string before the undefined-gate check
  runs. Reorder so the lookup failure is reported first.
- Add a defensive `try/except (AttributeError, TypeError)` around the visitor's top-level
  dispatch that re-raises as `ValidationError` with the statement span. This converts any
  remaining internal slips in the tree into the documented error type — a cheap safety net
  beyond this specific fix.
- Tests: `tests/qasm3/test_validation.py` — assert `ValidationError` (not `AttributeError`)
  for a bare binary expression, a bare function call, a bare identifier, and an undefined gate
  with zero and with one qubit operand.

danielgaskins:
## Summary

- evaluate and discard standalone expressions instead of assuming every expression statement is a function call
- preserve operations produced by subroutine and extern calls
- apply the same behavior inside OpenPulse calibration blocks while retaining specialized handling for pulse functions
- validate unknown gate names before their operands so the error identifies the gate

## Testing

- `pytest -q tests`
- `pytest -q --cov=pyqasm --cov-report=term tests`
- `isort --check-only src tests bin examples`
- `black --check --target-version py310 src tests bin examples`
- `pylint src tests bin examples`
- `mypy src examples`
- `qbraid admin headers src tests bin examples --skip=src/pyqasm/_version.py --type=apache`
- `sphinx-build -W -b html docs /tmp/pyqasm-388-docs`

Closes #388


src/pyqasm/visitor.py
```diff
@@ -3527,12 +3554,7 @@ def visit_statement(
 
         visitor_function = self._visit_map.get(type(statement))
         if visitor_function:
-            if isinstance(statement, qasm3_ast.ExpressionStatement):
-                # these return a tuple of return value and list of statements
-                _, ret_stmts = visitor_function(statement)  # type: ignore[operator]
-                result.extend(ret_stmts)
-            else:
-                result.extend(visitor_function(statement))  # type: ignore[operator]
+            result.extend(visitor_function(statement))  # type: ignore[operator]
```
ryanhill1: **N2 — #388 also asked for a defensive `try/except` here**, around the top-level dispatch, re-raising `AttributeError`/`TypeError` as `ValidationError` with the statement span — described in the issue as "a cheap safety net beyond this specific fix".

I think you're right to have left it out: a blanket catch at the dispatch would convert genuine internal bugs into validation errors and bury them, which is a bad trade for a net that my fuzzing suggests is already empty. No change requested.

The only issue is that the PR says `Closes #388`, so that item will silently close as done. Could you either add a line to the description saying it was deliberately declined and why, or drop the auto-close so the item can be judged separately?

src/pyqasm/pulse/visitor.py
```diff
@@ -773,6 +773,40 @@ def _visit_function_call(  # pylint: disable=too-many-branches, too-many-stateme
...
+        value.
+
+        Args:
+            statement (ExpressionStatement): The expression statement to visit.
+
+        Returns:
+            list[Statement]: Statements produced while evaluating the expression.
+        """
+        expression = statement.expression
+        pulse_functions = {
```
ryanhill1: **N3 — this list now has to stay in sync with `_visit_function_call`.**

That method special-cases eight names; four of them (`get_phase`, `get_frequency`, `newframe`, `play`) aren't in any of the three maps, so they're spelled out again here. I checked the current set is complete — `set_phase`, `set_frequency`, `shift_phase` and `shift_frequency` all come in via `OPENPULSE_FRAME_FUNCTION_MAP` — so there's no bug today.

But a ninth added to `_visit_function_call` and not here would silently reroute to the generic evaluator. Existing tests would catch it loudly, so this is maintainability only. Could you pull the four literals into a module-level constant that both sites use, or failing that leave a comment here pointing at `_visit_function_call`?

ryanhill1 (commented):
Thanks for this — it's a clean fix, and I confirmed it against every repro in #388: the two spec examples (`a * b;`, `a + b;`) and `g1;` all went from `AttributeError` to working or to a proper `ValidationError`, and `g1 q1;` with an undeclared operand now names the gate rather than the qubit. I also fuzzed 27 expression-statement forms (literals, unary/binary, index, slice, cast, ternary, `measure`, string, duration) — no non-`ValidationError` escapes. Suite, pylint, mypy, black and isort all green locally, and CI is green across the full wheel matrix.

Three minor things. **N1** is here; **N2** and **N3** are inline.

**N1 — please add a CHANGELOG entry** under `Fixed`, referencing this PR #425. This is user-facing twice over: the exception type changes from `AttributeError` to `ValidationError`, and undeclared-gate diagnostics now quote the original source line instead of the expanded one (which is why `tests/qasm3/resources/gates.py:397` had to change).

danielgaskins:
Thanks, those are good points. I added the changelog entry and a note beside the OpenPulse function names pointing to `_visit_function_call`. I also clarified in the PR description why I left out the broad exception catch. The new test parameters have types and argument docs now. The branch is updated from main, and the full suite passes.

ryanhill1 (dismissed):
Looks good from my end. I'll let @TheGupta2012 give final sign off before merge.

src/pyqasm/visitor.py
```diff
@@ -1638,6 +1638,15 @@ def _visit_generic_gate_operation(  # pylint: disable=too-many-branches, too-man
             )
             return stmts  # type: ignore
 
+        if (
```
TheGupta2012: Could you explain a little why this change is needed here? Just want to understand this through an example

TheGupta2012 (commented):
Thanks @danielgaskins for this change, apologies that I've been a little slow to review your PRs. Thanks for keeping up with the good work! 

Besides one clarification, the changes lgtm.

src/pyqasm/visitor.py
danielgaskins: Sure. With `unknown missing_qubit;`, the old path checks the qubit first, so it reports `missing_qubit` without ever identifying the invalid gate. This check looks up `unknown` first and reports the gate name at the original source line. It skips custom and black-box gates, so their handling stays the same. The new tests cover both this case and an unknown gate with a declared qubit.

